PR #3, opened 2017-05-02
Fixes reference stack recommendation to pick the best reference stack

miteshvp:


bkabrda:
Please rebase, we've been doing changes to not point at the old `baytemp` org, but you sent the PR too soon and Jenkins still sees the old links and fails because of that. Rebasing on top of latest master should fix this.

bkabrda:
Otherwise this LGTM, so once Jenkins run passes, I can merge.

bkabrda:
Hmm, ok, I see that the problem is elsewhere and I opened https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/8 to fix it. This looks fine, so I'm merging.

PR #7, opened 2017-05-04
Improve worker logging

fridex:


pkajaba:
[test this]

fridex:
> LGTM, can this be merged ?

There is unclosed discussion on ML regarding this. I will refresh it.

BTW this would also address log size issues.

fridex:
So based on internal discussion, we will have to slightly change this. Closing this, but the new PR will be in many aspects based on this.

PR #16, opened 2017-05-19
[openshift] 9 replicas

msrb:


msrb:
We need this so the deployment won't get accidentally scaled down to 1 in dsaas. Merging.

PR #22, opened 2017-05-22
Create index on ecosystem name, not on its URL

fridex:


pkajaba:
Just to be sure what you want to accomplish. Do you want to create index just for the name of package?

fridex:
Yes, only name of the package should be indexed (unique index).

fridex:
@pkajaba PTAL

pkajaba:
LGTM, thx

PR #23, opened 2017-05-23
Transform package names for case insensitive ecosystems

fridex:


fridex:
We would probably like to utilize this with functions like `MavenCoordinates.normalize_str()` to centralize sanity checks and transformations on package names that enter to our system.

cucoslib/workers/init.py
```diff
@@ -15,6 +15,9 @@ def execute(self, arguments):
         self._strict_assert(arguments.get('version'))
         self._strict_assert(arguments.get('ecosystem'))
 
+        # make sure we store package name based on ecosystem package naming case sensitivity
+        arguments['name'] = case_sensitivity_transform(arguments['name'])
```
tuxdna: `case_sensitivity_transform` takes two arguments `(ecosystem, name)` , but here only one argument is provided `arguments['name']`. Is this correct ?

fridex: Thanks! Nice catch. I was too fast with `git push` :)

fridex:
PTAL @tuxdna

tuxdna:
LGTM, merging now.

PR #24, opened 2017-05-23
[openshift] Fix "[pos 376]: json: decNum: got first char '$'"

msrb:


msrb:
Typo, merging myself.

PR #25, opened 2017-05-24
Average of averages is not average

fridex:
Remove confusing metric.


pkajaba:
LGTM! Thanks, It did not make sense to have average of averges

PR #28, opened 2017-05-29
[openshift] Lower memory limit to 1GB

msrb:
So we can scale to 30 workers


msrb:
We need this in order to scale workers to 30.

PR #27, opened 2017-05-29
Introduce package-specific flows

fridex:

PR #31, opened 2017-06-01
Add flow and tasks for processing most starred projects on GitHub

msrb:


cucoslib/dispatcher/nodes.yml
```diff
@@ -215,6 +215,15 @@
       max_retry: 0
       queue: '{DEPLOYMENT_PREFIX}_{WORKER_ADMINISTRATION_REGION}_GitReadmeCollector_v0'
       storage: 'S3Readme'
+    - name: 'GitHubManifestMetadataResultCollector'
+      import: 'cucoslib.workers'
+      max_retry: 0
+      queue: '{DEPLOYMENT_PREFIX}_{WORKER_ADMINISTRATION_REGION}_GitHubManifestMetadataResultCollector_v0'
+    - name: 'GitHubManifestMetadataInit'
```
fridex: Let's be unified with naming - `InitGitHubManifestMetadata`

cucoslib/dispatcher/flows/githubManifestMetadataFlow.yml
```diff
@@ -0,0 +1,30 @@
...
+          to: 'metadata'
+        - from: 'metadata'
+          to: 'dependency_snapshot'
+        - from: 'dependency_snapshot'
+          to: 'bayesianFlow'
+          foreach:
+            import: 'cucoslib.dispatcher.foreach'
+            function: 'iter_dependencies_analysis'
+            propagate_result: true
+          condition:
```
fridex: I believe these conditions are not necessary.

cucoslib/dispatcher/nodes.yml
```diff
@@ -325,6 +335,16 @@
         encryption: false
         versioned: true
 
+    - name: 'S3GitHubManifestMetadata'
+      import: 'cucoslib.storages'
+      configuration:
+        aws_access_key_id: 'GNV3SAHAHA3DOT99GQII'
+        aws_secret_access_key: 'ZmvMwngonaDK5ymlCd6ptaalDdJsCn3aSSxASPaZ'
```
fridex: Based on https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/27 please replace this with `<<: *configurationS3` to avoid copy-paste errors.

cucoslib/workers/gh_metadata_init.py
```diff
@@ -0,0 +1,19 @@
+import datetime
+from cucoslib.base import BaseTask
+from cucoslib.models import Analysis
+
+
+class GitHubManifestMetadataInit(BaseTask):
+    def execute(self, arguments):
+        self._strict_assert(arguments.get('git_repo_url'))
```
fridex: Could we unify `git_repo_url` with `url`?

cucoslib/workers/gh_metadata_result_collector.py
```diff
@@ -0,0 +1,32 @@
...
+from cucoslib.base import BaseTask
+from cucoslib.utils import get_analysis_by_id
+
+
+class GitHubManifestMetadataResultCollector(BaseTask):
+    """
+    Collect all results that were computed, upload them to S3 and store version reference to results in WorkerResult
+    """
+
+    def __init__(self, *args, **kwargs):
```
fridex: Not needed, the constructor of parent task gets called implicitly.

cucoslib/workers/gh_metadata_result_collector.py
```diff
@@ -0,0 +1,32 @@
...
+        super().__init__(*args, **kwargs)
+
+    def run(self, arguments):
+        self._strict_assert(arguments.get('ecosystem'))
+        self._strict_assert(arguments.get('repo_name'))
```
fridex: Why there is `repo_name`? Based on the job handler it is basically a duplicity of the repo URL.

cucoslib/workers/mercator.py
```diff
@@ -157,6 +164,23 @@ def execute(self, arguments):
...
+        workdir = None
+        try:
+            workdir = tempfile.mkdtemp()
+            repo_url = arguments.get('git_repo_url')
+            Git.clone(repo_url, path=workdir, depth=str(1))
+            metadata = self.run_mercator(arguments, workdir)
+            if metadata.get('status', None) != 'success':
+                self.log.error('Mercator failed on %s', repo_url)
+                return None
+            return metadata
```
fridex: Here and in the case above, the temporary dir won't get deleted.

msrb:
Thanks! I am going to fix the issues.

cucoslib/workers/gh_metadata_init.py
msrb: Could you please clarify? Do we have `url` somewhere already?

cucoslib/workers/gh_metadata_result_collector.py
msrb: good catch, thanks.

cucoslib/workers/gh_metadata_init.py
fridex: Not yet, but https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/27/files#diff-8ccf2d5f54c7573302ac2b0200a732edR255 will introduce it. Let's be consistent as we might want to do monitoring on these projects as well. So either I will introduce `git_repo_url` or you make the change. It's up to you :)

cucoslib/workers/gh_metadata_result_collector.py
msrb: `repo_name` should be just `<username>/<repo>` portion of the URL. I later store results in S3 under `username:repo/<result>.json`. I could pass just URL and then extract `repo_path` from there, when needed. But it was easier this way :)

cucoslib/workers/mercator.py
msrb: It's inside `try-finally` so it should be called in both cases.

cucoslib/workers/gh_metadata_result_collector.py
fridex: Yeah, I saw it. Maybe we would like to unify storing by parsing url as with the current approach there could be created for example `www.github.com` and `github.com` S3 "folder" - not sure about the data from jobs though (note that the flow can be scheduled even manually).

cucoslib/workers/mercator.py
fridex: +1, missed it.

cucoslib/workers/gh_metadata_result_collector.py
fridex: Now I see you do substitution. OK.

cucoslib/dispatcher/flows/githubManifestMetadataFlow.yml
msrb: You sure about this? When I removed those it looked like I wasn't able to control whether I want to analyze transitive dependencies or not.

msrb:
Updated so raw manifest files are also stored in S3.

cucoslib/dispatcher/flows/githubManifestMetadataFlow.yml
fridex: What confused me is the fact that there are no analyses in this flow. But given the fact we want to reuse `iter_dependencies_analysis` that decrements `recursive_limit`, so we might do it this way (zero means no analyses at all).

msrb:
Rebased, PTAL.

PR #38, opened 2017-06-06
Fix "TypeError: get_analysis_by_id() takes 2 positional arguments but…

msrb:
… 4 were given"


msrb:
Fixes bug introduced by me while rebasing my GH-most-starred PR.

PR #39, opened 2017-06-07
Fix several errors reported by PEP8 checker

tisnik:
This changeset just fixes several errors reported by the PEP8 checker. It does not change the program logic at all, just makes sure we'd follow standards ;)


tisnik:
wow that was fast! TYVM

PR #40, opened 2017-06-07
[PackagePostgres] add get_analysis_count(ecosystem, package)

jpopelka:


cucoslib/storages/package_postgres.py
```diff
@@ -45,3 +45,21 @@ def get_analysis_by_id(self, analysis_id):
...
+
+    def get_analysis_count(self, ecosystem, package):
+        """Get count of previously scheduled analyses for given ecosystem-package.
+
+        :param ecosystem: str, Ecosystem name
+        :param package: str, Package name
+        :return: analysis count
+        """
+        if ecosystem == 'maven':
+            package = MavenCoordinates.normalize_str(package)
```
fridex: We will probably need to do case-sensitivity normalization for Python package names, but we can do that later.

PR #50, opened 2017-06-09
Make S3GitHubManifestMetadata.get_object_key_path() public

msrb:


msrb:
cc @fridex @jpopelka 

I need this deployed so we can analyze stuff over the weekend, therefore merging myself. Feel free to review. I will address all comments/issues.

PR #56, opened 2017-06-14
Utilize ScanCode toolkit in LicenseCheckTask (source_licenses)

jpopelka:


fridex:
With this we would probably like to reschedule license tasks to get new output for all artefacts that we have already analysed.

miteshvp:
Yes, makes sense. I guess we may also have to run github package specific flow, correct?

fridex:
> ... run github package specific flow, correct?

This is not related to this PR, but we eventually would like to do some one time job for github_details related transition.

pombredanne:
@jpopelka very nice! I will be on deck if there are any bug or issues!

PR #60, opened 2017-06-16
Ingest unknown dependencies found during stack analysis

samuzzal-choudhury:


fridex:
Why this is not done in the YAML configuration?

samuzzal-choudhury:
@fridex educate me.

fridex:
> @fridex educate me how.

I think we would like to keep this configurable as it grants us more control and keeps tasks as simple as possible.

There will be needed a `foreach` function as described in [1]. Take a look at existing implementations in `cucoslib/dispatcher/foreach.py` and `bayesianAnalysisFlow` config file. A lot could be reused from this PR, so it shouldn't take that much time. Reach out to me in case of any questions.

Thanks Sam!

[1] https://selinon.readthedocs.io/en/latest/patterns.html#variable-number-of-child-tasks-or-sub-flows

PR #64, opened 2017-06-19
Remove duplicate paths in license scanner output

jpopelka:
If a license is found more than once in a file, then 'paths' contains the file more than one.
Remove the duplicates by using set.


pombredanne:
@jpopelka tell me if you think this is something weird in that I could be fix upstream or if this is something specific to your use case.

jpopelka:
@pombredanne no worries, upstream is fine

jpopelka:
[test this]

PR #83, opened 2017-07-03
Pin version of scancode, because upsteam is broken. 

pkajaba:
This is workaround until upsteam is fixed
I published fix in upstream repository: https://github.com/nexB/scancode-toolkit/pull/671


pkajaba:
Fixed in upstream

PR #85, opened 2017-07-04
Run also githuber in githubManifestMetadataFlow

msrb:


f8a_worker/dispatcher/nodes.yml
```diff
@@ -247,6 +247,12 @@
       max_retry: 0
       queue: '{DEPLOYMENT_PREFIX}_{WORKER_ADMINISTRATION_REGION}_InitGitHubManifestMetadata_v0'
       storage: 'BayesianPostgres'
+    - name: 'gh_most_starred_github_details'
```
msrb: We are using different storage here, that's why the different name.

pkajaba:
Thanks, looks good.

PR #87, opened 2017-07-04
Fix PypiSolver

jpopelka:
Failing `test_pypi` is expected - to show that `PypiSolver` is wrong and needs to be fixed.


pkajaba:
LGTM, Thanks, Jirka!

PR #88, opened 2017-07-06
WIP: Common bookkeeping for Stack and Component Analysis API

samuzzal-choudhury:


f8a_worker/storages/postgres.py
```diff
@@ -89,7 +90,7 @@ def get_worker_id_count(self, worker_id):
         return PostgresBase.session.query(WorkerResult).filter(WorkerResult.worker_id == worker_id).count()
 
     def get_analysis_by_id(self, analysis_id):
-        """Get result of previously scheduled analysis
+        """Get result of previously scheduled analysis5
```
pkajaba: This is probably typo

f8a_worker/storages/postgres.py
```diff
@@ -100,3 +101,30 @@ def get_analysis_by_id(self, analysis_id):
...
+            origin = data.get('origin', None),
+            request = data.get('request', None),
+            team = data.get('team', None),
+            recommendation = data.get('recommendation', None)
+        )
+
+        PostgresBase.session.add(req)
+        PostgresBase.session.commit()
+
+        return True
```
pkajaba: is this 'True' really required?

f8a_worker/workers/bookkeeper.py
```diff
@@ -0,0 +1,21 @@
...
+    # we don't want to add `_audit` etc into the manifest submitted
+    add_audit_info = False
+
+    def execute(self, arguments):
+        self._strict_assert(arguments.get('external_request_id'))
+        self._strict_assert(arguments.get('data'))
+
+        self.log.info("Request id = %s" % arguments.get('external_request_id'))
+
+        postgres = StoragePool.get_connected_storage('BayesianPostgres')
```
fridex: I would rather suggest to create a new adapter besides `BayesianPostgres` and place the storing logic there. Main benefits:

1. We do not mix analyses storing logic with the bookkeeping logic
2. If we decide to introduce a new RDS instance just for data like this, we will just pass different connection string (will help us scale).
3. It will more fit the current design where we configure tasks using YAML config files.

Anyway 1. is more reasonable in short term.

samuzzal-choudhury:
A cleaner PR https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/114 has been opened for the same feature. Hence, closing this.

PR #93, opened 2017-07-11
Introduce GitHub statistics task

fridex:


fridex:
We could introduce more statistics, just proposed idea.

f8a_worker/workers/git_stats.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        dir = mkdtemp()
+
+        try:
+            git = Git.clone(url, dir)
+            # nice notebook to check at:
+            #   http://nbviewer.jupyter.org/github/tarmstrong/code-analysis/blob/master/IPythonReviewTime.ipynb
+            log = git.log()
+        finally:
+            shutil.rmtree(dir, ignore_errors=True)
```
jpopelka: You can use [tempdir](https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/f8a_worker/utils.py#L141):

```python
with tempdir as dir:
    git = Git.clone(url, dir)
    log = git.log()
```

f8a_worker/workers/git_stats.py
```diff
@@ -0,0 +1,159 @@
...
+
+        :param trend: numerical representation of thrend
+        :return: textual representation of trend
+        """
+        if trend > 3:
+            return 'increasing'
+        elif trend > -3:
+            return 'calm'
+        else:
+            return 'decreasing'
```
jpopelka: Isn't the -3 to 3 boundary for 'calm' quite big ? I'd expect most of the trends to be between -3 to 3.

f8a_worker/workers/git_stats.py
```diff
@@ -0,0 +1,159 @@
...
+            ('overall', master_log, master_log[-1]['author']['date'] if master_log else None),
+            ('year', last_year_log, last_year_timestamp),
+            ('month', last_month_log, last_month_timestamp)
+        )
+        for key, log, starting_date in records:
+            master_stats[key] = self._compute_stats(log, starting_date)
+
+        orgs = self._get_orgs(master_log)
+
+        return {'summary': [], 'status': 'success', 'details': {'master': master_stats, 'organizations': orgs}}
```
jpopelka: Why do we have the organizations separately ? What about having it in `master_stats` ?

fridex: I wanted to distinguish time-based stats present in `master_stats` from the "generic ones" like organizations.

f8a_worker/workers/git_stats.py
fridex: Yeah, good catch. I think we should lower the boundary. Any suggestions?

f8a_worker/workers/git_stats.py
fridex: Thanks, I missed that helper. Changed accordingly.

f8a_worker/workers/git_stats.py
jpopelka: I guess you've been playing with that so you probably have better idea, but for me something like <-1,1> (or even smaller interval) would sound reasonable.

It'd also be nice to have few simple tests covering these methods.

jpopelka: Also, first time I saw the condition I had problems understanding it. For me something like

```python
elif trend < -3:
    return 'decreasing'
else:
    return 'calm'
```

would be more 'parsable'.

PR #100, opened 2017-07-12
[DataNormalizer] treat download-url as code_repository

### LINKED ISSUE #97: Capture download_url from setup.py
Package level flows fail if the link to GitHub repo is stored under `download_url` instead of `url`. An example could be httpie: https://github.com/jakubroztocil/httpie/blob/master/setup.py#L79

To solve this, we should propagate only GitHub urls in the `url` field.

jpopelka:
Fixes #97

### DIFF
--- a/f8a_worker/data_normalizer.py
+++ b/f8a_worker/data_normalizer.py
@@ -242,20 +242,25 @@ class DataNormalizer(object):
         return keywords
 
     def _handle_python(self, data):
-        "Handle Python package (setup.py) analysis data"
+        """ setup.py """
+        if 'error' in data:
+            # mercator by default (MERCATOR_INTERPRET_SETUP_PY=false) doesn't interpret setup.py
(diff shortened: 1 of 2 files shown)


f8a_worker/data_normalizer.py
```diff
@@ -242,20 +242,25 @@ def _split_keywords(keywords, separator=','):
         return keywords
 
     def _handle_python(self, data):
-        "Handle Python package (setup.py) analysis data"
+        """ setup.py """
```
pkajaba: Just a little nitpicking - this is not related to this PR that much, so it would be nice to do it as a separate commit.

pkajaba:
Thanks, LGTM

f8a_worker/data_normalizer.py
jpopelka: ok, next time :-)

PR #102, opened 2017-07-13
Crude workaround for amqp failures

miteshvp:


fridex:
This shouldn't occur in our version as we have pinned version to 2.1.4. Based on https://github.com/celery/celery/issues/4024#issuecomment-314864073

fridex:
OK, I think we get newer one when installing selinon.

msrb:
@fridex we install kombu from fork and I think that is where we pull the newer version.

msrb:
Hmm, indeed, we install selinon with pip's `-U` flag. So it's possible we will pull the newer version there.

PR #103, opened 2017-07-13
Delete unused import

msrb:


msrb:
Dummy change to force CI to pick up change from my kombu fork.

PR #104, opened 2017-07-13
Only reinstall selinon, but don't touch already installed dependencies

msrb:


msrb:
cc @miteshvp @fridex 

Temp. quick workaround before we figure out how to solve this dependency problem properly. I need to test it first though.

msrb:
I was able to build the image locally, with `--no-cache` and the `amqp` version is 2.14 - as expected. I think this can be merged.

PR #109, opened 2017-07-13
Update selinon and selinonlib to latest versions

pkajaba:


msrb:
@fridex has to give thumbs up on this one :)

fridex:
I was about to merge this - there are no significant changes that would affect functionality. Let's keep this open and I will merge this after demos. I want to release a new version anyway and the best way to test it is to use selinon/selinonlib in this project :)

pkajaba:
@fridex it's your call :-) However, I agree with you

PR #115, opened 2017-07-17
Sentiment scorer for stack analysis

harjinder-hari:


f8a_worker/workers/sentiment_scorer.py
```diff
@@ -0,0 +1,324 @@
...
+
+        sentiment_output = {}
+        arg_instances = []
+        for result in aggregated['result']:
+            resolved = result['details'][0]['_resolved']
+            ecosystem = result['details'][0]['ecosystem']
+            for elm in resolved:
+                arg_instances.append((ecosystem, elm['package']))
+
+        sentiment_results = Parallel(n_jobs=4, verbose=1, backend="threading")\
```
msrb: I think this won't work as expected, but I can be wrong. In any case I am not going to block PR on this as this is just an optimization.

https://pythonhosted.org/joblib/parallel.html#using-the-threading-backend


> If you know that the function you are calling is based on a compiled extension that releases the Python Global Interpreter Lock (GIL) during most of its computation then it might be more efficient to use threads instead of Python processes as concurrent workers. For instance this is the case if you write the CPU intensive part of your code inside a with nogil block of a Cython function.

`sentiment_analysis()` will, in my opinion, hold the GIL and won't release it.

requirements.txt
```diff
@@ -4,52 +4,104 @@
...
+proto-google-cloud-pubsub-v1[grpc]==0.15.4  # via gapic-google-cloud-pubsub-v1
+proto-google-cloud-spanner-admin-database-v1[grpc]==0.15.3  # via gapic-google-cloud-spanner-admin-database-v1
+proto-google-cloud-spanner-admin-instance-v1[grpc]==0.15.3  # via gapic-google-cloud-spanner-admin-instance-v1
+proto-google-cloud-spanner-v1[grpc]==0.15.3  # via gapic-google-cloud-spanner-v1
+proto-google-cloud-speech-v1[grpc]==0.15.3  # via gapic-google-cloud-speech-v1
+protobuf==3.3.0           # via google-cloud-core, google-gax, googleapis-common-protos, grpcio
+psycopg2==2.7.1
+pyasn1-modules==0.0.9     # via google-auth, google-auth-httplib2, oauth2client
+pyasn1==0.2.3             # via google-auth, google-auth-httplib2, oauth2client, pyasn1-modules, rsa
+pygithub==1.35
```
msrb: Our downstream patch doesn't apply to this newer version.

f8a_worker/workers/sentiment_scorer.py
```diff
@@ -0,0 +1,324 @@
...
+
+
+class GoogleSentimentAnalyzer(SentimentAnalyzer):
+    def __init__(self, key_file_path):
+        os.environ['GOOGLE_APPLICATION_CREDENTIALS'] = key_file_path
+        self.lang_client = language.Client()
+
+    def analyze_sentiment(self, text):
+        start_time = time.time()
+        print("Started analyzing sentiment")
```
msrb: We could use proper logging here, but given time constrains, this will do just fine for now.

requirements.txt
msrb: Oh, actually it's `PyGithub` vs `pygithub` problem again...

requirements.txt
```diff
@@ -4,52 +4,104 @@
...
 anymarkup==0.7.0
-beautifulsoup4==4.5.3
-BigQuery-Python==1.10.0
+beautifulsoup4==4.6.0
+bigquery-python==1.12.0
 billiard==3.5.0.2         # via celery
 boto3==1.4.4
-boto==2.45.0
-botocore==1.5.0           # via boto3, s3transfer
+boto==2.48.0
```
msrb: Updating boto makes our downstream patch unnecessary - it fails to apply now, we should drop it.

msrb:
Hmm, I must be doing something wrong, but I can't get it working locally. I am going to trust @harjinder-hari that he tested the changes thoroughly :)

msrb:
The PR doesn't seem to break anything (except downstream patches) so I am going to merge this to unblock @harjinder-hari. Let's see if it works in dev/stage.

Dockerfile
```diff
@@ -117,3 +117,5 @@ COPY hack/patches/* /tmp/install_deps/patches/
 COPY hack/apply_patches.sh /tmp/install_deps/
 # Apply patches here to be able to patch selinon as well
 RUN /tmp/install_deps/apply_patches.sh
+
+RUN pip3 uninstall -y protobuf && pip3 install packaging appdirs && pip3 install --upgrade --no-binary :all: protobuf==3.3.0
```
harjinder-hari: We encountered a very weird Import Error with Google Cloud lib. Even if the lib is installed properly and we can see manually that it is available in import-path, but still the program throws Import Error for "protobuf" !

I discussed it in matter-most and Saket/Slavek shared the above work-around.
Reference: https://chat.openshift.io/rh-devtools/pl/5e6sbcsyz3fuxfw1zs4n97hwww

f8a_worker/workers/sentiment_scorer.py
```diff
@@ -0,0 +1,324 @@
...
+            toList();
+            """.format(pkg_name=pkg_name,
+                       score=sentiment_details['score'],
+                       magnitude=sentiment_details['magnitude'],
+                       last_updated=sentiment_details['last_updated'])
+        self.__execute_query(query)
+        return
+
+
+def write_key_file(local_path):
```
harjinder-hari: Sure Michal, I have opened a new issue #129 to keep track of this.

f8a_worker/workers/sentiment_scorer.py
harjinder-hari: Sure Michal, I have opened a new issue #129 to keep track of this.

requirements.txt
harjinder-hari: Sure Michal, I have opened a new issue #129 to keep track of this.

f8a_worker/workers/sentiment_scorer.py
harjinder-hari: Sure Michal, I have opened a new issue #129 to keep track of this.

requirements.txt
harjinder-hari: Sure Michal, I have opened a new issue #129 to keep track of this.

PR #122, opened 2017-07-17
Do graph imports in package-level flows

fridex:


miteshvp:
>We just remove version info in the request to graph importer. Will it be OK?

@fridex - I am not sure if I completely follow you, but, graph_importer requires EPV. It then picks up Package and Version from respective buckets as part of config paramters

miteshvp:
LGTM otherwise. Just one query as above.

fridex:
> @fridex - I am not sure if I completely follow you, but, graph_importer requires EPV. It then picks up Package and Version from respective buckets as part of config paramters

I'm not sure if that is what we want. We now have package-version-level and package-level flows. We would like to run graph imports in each flow separately after all tasks finish. This way we can refresh github details without running our main flow.

fridex:
Closing this as graph importer task will be called from the main flow. However we will need similar functionality in the future when we will want to run only tasks in package level flow.

PR #128, opened 2017-07-17
[openshift/template.yaml] increase memory limit to 1200Mi due to scancode

### LINKED ISSUE #127: 1024Mi memory limit is not enough for scancode
Current limit for worker is [1024Mi](https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/openshift/template.yaml#L227)

`docker stats` on locally running worker container reveals that during license scan the memory usage spikes to something around 1024Mi, which is probably the reason why on Openshift many license scans fail with no obvious reason.

jpopelka:
Fixes #127

### DIFF
--- a/openshift/template.yaml
+++ b/openshift/template.yaml
@@ -224,7 +224,7 @@ objects:
               memory: "512Mi"
               cpu: "300m"
             limits:
-              memory: "1024Mi"
+              memory: "1200Mi"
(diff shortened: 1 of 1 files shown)


jpopelka:
[test this]

msrb:
LGTM, thanks 😉

Issue #141, opened 2017-07-19
User Profile storage on S3 failed


samuzzal-choudhury:
The bookkeepertask fails when tries to store user profile on S3 with the following error -

```
  File "/usr/lib/python3.4/site-packages/botocore/validate.py", line 291, in serialize_to_request
    raise ParamValidationError(report=report.generate_report())
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid type for parameter ServerSideEncryption, value: True, type: <class 'bool'>, valid types: <class 'str'>
```
### Reason:
This happens because of the wrong encryption method set in the nodes.yml file for S3UserProfileStore adapter.

### Remedy
Make S3UserProfileStore encryption false which is in sync with other S3 adapters.

samuzzal-choudhury:
aws:kms is the appropriate choice for encryption

jpopelka:
Fixed by #142

PR #142, opened 2017-07-19
Added proper server side encryption for user profile bucket

samuzzal-choudhury:
Fixes issue https://github.com/fabric8-analytics/fabric8-analytics-worker/issues/141

### DIFF
--- a/f8a_worker/dispatcher/nodes.yml
+++ b/f8a_worker/dispatcher/nodes.yml
@@ -412,7 +412,7 @@
         <<: *configurationS3
         bucket_name: '{DEPLOYMENT_PREFIX}-janus-json-data'
         region_name: 'us-east-1'
-        encryption: true
+        encryption: false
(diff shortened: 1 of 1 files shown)


samuzzal-choudhury:
Before doing this change, I tried setting the encryption to 'aes256' which also did not work. I think it will need a broader discussion for whether or not we require encryption for our S3 stores and also the method to be used.

Until then, IMHO we should set encryption false to S3UserProfileStore which will let BookkeeperTask complete.

msrb:
@samuzzal-choudhury I can see there is `aws:kms` encryption set for (almost?) all S3 storages. That one doesn't work for you?

samuzzal-choudhury:
@msrb thanks for pointing me to aws:kms. Done the change and it's working fine. Please review and merge if you think that the changes are good. CC: @fridex

PR #160, opened 2017-07-28
remove unrequired print statement

SaketJNU:
Unnecessary print() statement in sentiment_scorer.py file is removed.


msrb:
LGTM. Just a note: since we are removing the print statements here and there is no other logging in the module, it's possible that we will have hard time debugging problems in future. But logging can be added in another PR.

SaketJNU:
Thanks @msrb, will create a new PR for logging facility.

PR #164, opened 2017-07-31
Enable selective syncs for graph imports

fridex:


fridex:
This solution is only partial - if some task fails in the analysis flows (package or package-version-level), syncs will not happen. Solutions:

1. Implement https://github.com/selinon/selinon/issues/24 so graph imports can be run on failures as well.
2. Use RDS to query for finished tasks to be synced.

IMHO 2. is easier, 1. is more "pretty" solution.

fridex:
Related: https://github.com/fabric8-analytics/fabric8-analytics-data-model/pull/32

fridex:
> IMHO 2. is easier, 1. is more "pretty" solution.

Chosen 2., we can switch to 1. later.

fridex:
@miteshvp PTAL - is data importer returning HTTP status code 200 when there is submitted EPV to sync?

BTW with this PR we will also be able to force graph sync of package data.

fridex:
@miteshvp also I don't know how exactly the data importer distinguishes between package-level and package-version level data. Is supplying `version: None` correct?

miteshvp:
> is data importer returning HTTP status code 200 when there is submitted EPV to sync?

Yes, it returns 200 when there is EPV synced to graph.

> also I don't know how exactly the data importer distinguishes between package-level and package-version level data. Is supplying version: None correct?

You can either provide version with None or Ecosystem:Package with no version as well. data importer will handle the scenario. e.g if you want to sync Package,Version for package 1 and just Package for a different package, you can provide something like
```
{
	
	"select_ingest":["libraries_io", "security_issues", "github_details", "source_licenses"],
	"package_list": 
		[
			{
				"version":"14.7.0",
				"name": "pyzmq", 
				"ecosystem": "pypi"
			},
                        {
				"name": "trollius", 
				"ecosystem": "pypi"
			}
		]
}
```

fridex:
@miteshvp thanks for clarification - in that case this PR is ready for review and to be merged.

f8a_worker/workers/graph_importer.py
```diff
@@ -4,24 +4,41 @@
...
 
-        ecosystem = arguments.get('ecosystem')
-        name = arguments.get('name')
-        version = arguments.get('version')
+        param = package_list
+        # If we force graph sync, sync all task results, otherwise only finished in this analysis run
+        if not arguments.get('force_graph_sync'):
+            # Tasks that need sync to graph start lowercase.
+            param = {
+                'select_list': [task_name
```
miteshvp: key name is `select_ingest`

PR #170, opened 2017-07-31
Remove hack/pip-requirements.txt

jpopelka:
I can't find what actually needs gofedlib.
If mercator-go, then the requirement should go into https://github.com/fabric8-analytics/mercator-go/blob/master/mercator.spec
If some worker, then it should go into requirements.in/txt


jpopelka:
PTAL Frido

PR #173, opened 2017-08-01
stack license and license outliers

rouseguy:
This PR calls the stack license API and returns stack license. If there are outliers with respect to license, it returns the outlier packages too.

### DIFF
--- a/f8a_worker/workers/stackaggregator_v2.py
+++ b/f8a_worker/workers/stackaggregator_v2.py
@@ -13,6 +13,7 @@ from f8a_worker.base import BaseTask
 from f8a_worker.graphutils import GREMLIN_SERVER_URL_REST, LICENSE_SCORING_URL_REST
 from f8a_worker.utils import get_session_retry
 
+import logging
 
 def extract_component_details(component):
     github_details = {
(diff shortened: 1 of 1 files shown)


f8a_worker/workers/stackaggregator_v2.py
```diff
@@ -13,6 +13,7 @@
 from f8a_worker.graphutils import GREMLIN_SERVER_URL_REST, LICENSE_SCORING_URL_REST
 from f8a_worker.utils import get_session_retry
 
+import logging
 
 def extract_component_details(component):
```
fridex: Is this function called from somewhere else? It would be more suitable to place it to correspondent class (`StackAggregatorV2Task` I believe) to encapsulate logic.

f8a_worker/workers/stackaggregator_v2.py
```diff
@@ -117,12 +118,34 @@ def aggregate_stack_data(stack, manifest_file, ecosystem, deps):
         "packages": license_score_list
     }
 
+    _logger = logging.getLogger(__name__)
```
fridex: Please use instantiated logger in `BaseTask` (available under `self.log`). We configure logging at one place so we are consistent.

f8a_worker/workers/stackaggregator_v2.py
```diff
@@ -117,12 +118,34 @@ def aggregate_stack_data(stack, manifest_file, ecosystem, deps):
...
+        
+        stack_license_exists = resp.get("stack_license_exists") 
+        package_with_unknown_license_exists = resp.get("package_with_unknown_license_exists")
+        stack_license_conflict_exists = resp.get("stack_license_conflict_exists")
+        package_license_conflict_exists = resp.get("package_license_conflict_exists")
+        packages_with_license_conflict = resp.get("packages_with_license_conflict")
+        recommended_stack_licenses = resp.get("recommended_stack_licenses")
+        stack_license_conflict_packages = resp.get("stack_license_conflict_packages")
+        stack_license_outlier_packages = resp.get("stack_license_outlier_packages")
+
```
fridex: Probably not suitable to try-except block.

f8a_worker/workers/stackaggregator_v2.py
```diff
@@ -117,12 +118,34 @@ def aggregate_stack_data(stack, manifest_file, ecosystem, deps):
...
     except:
-        stack_license = [None]
+        _logger.error("Error getting data from Stack License API")
```
fridex: `self.log.exception("Error getting data from Stack License API")` to capture exception details as well.

f8a_worker/workers/stackaggregator_v2.py
```diff
@@ -117,12 +118,34 @@ def aggregate_stack_data(stack, manifest_file, ecosystem, deps):
...
+
+        #Construct default response if API has issues
```
fridex: Just a design question - do we want to have this default response here? What about reporting error and letting user to try again?

rouseguy:
This PR will be replaced with another one by @harjinder-hari - which uses the latest version of json output

PR #176, opened 2017-08-01
Enforce timeout in license worker

msrb:


f8a_worker/workers/license.py
```diff
@@ -88,12 +90,14 @@ def execute(self, arguments):
...
             with username():
-                output = TimedCommand.get_command_output(command,
-                                                         graceful=False,
-                                                         is_json=True,
-                                                         timeout=1200)
+                tc = TimedCommand(command)
+                status, output, error = tc.run(is_json=True, timeout=1200)
+                if status != 0:
+                    self.log.error(error)
+                    raise TaskError("Error (%s) during running command %s: %r" % (str(status), command, output))
```
fridex: `selinon.FatalTaskError` would be more suitable here

msrb: Fixed, thanks ;)

Issue #175, opened 2017-08-01
[anitya_worker] dependency on digests


msrb:
Our anitya worker has, from some reason, dependency on results from digests worker. This is kinda weird and unexpected, IMO. We should investigate.

fridex:
It's used for computed artefact hashes in https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/f8a_worker/workers/anitya.py#L93. We could adjust digester to compute only digests of artefacts (only artefacts) to keep this functionality alive.

msrb:
If the only purpose of anitya is to monitor upstream repositories for new releases, then I don't understand why we care about downstream rpms and hashes. I am sure the code is there for a reason, it's just not obvious why.

jpopelka:
Yeah, some parts of the [AnityaTask](https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/f8a_worker/workers/anitya.py#L21) should IMHO be actually in [DownstreamUsageTask](https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/f8a_worker/workers/downstream.py#L28), like this [_get_downstream_rpm_pkgs()](https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/f8a_worker/workers/anitya.py#L93) where we use the artifact hash to get the downstream package name from `DownstreamMapCache` (basically `hash -> downstream-package-name` mapping created from dist-git)

msrb:
Digester task has been re-enabled, but we now only compute hashes for the tarball itself.

jpopelka:
Anitya was removed in https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/491

PR #178, opened 2017-08-01
Re-enable digests worker again

msrb:


msrb:
Until https://github.com/fabric8-analytics/fabric8-analytics-worker/issues/175 is sorted out, we will continue collecting digests.

PR #179, opened 2017-08-01
Pick the correct pgm scoring

miteshvp:


f8a_worker/workers/recommender.py
```diff
@@ -514,9 +514,13 @@ class RecommendationV2Task(BaseTask):
 
     def call_pgm(self, payload):
         """Calls the PGM model with the normalized manifest information to get the relevant packages"""
-        pgm_url = PGM_URL_REST + "/api/v1/schemas/kronos_scoring"
         try:
-            if payload is not None:
+            # TODO remove hardcodedness for payloads with multiple ecosystems
+            if payload is not None and 'ecosystem' in payload[0]:
```
msrb: You could say `if payload and 'ecosystem' in payload[0]:` - that way you would catch scenario when `payload` variable is an empty array.

f8a_worker/workers/recommender.py
```diff
@@ -514,9 +514,13 @@ class RecommendationV2Task(BaseTask):
 
     def call_pgm(self, payload):
         """Calls the PGM model with the normalized manifest information to get the relevant packages"""
-        pgm_url = PGM_URL_REST + "/api/v1/schemas/kronos_scoring"
         try:
-            if payload is not None:
+            # TODO remove hardcodedness for payloads with multiple ecosystems
```
msrb: Not sure what this TODO means exactly :)

f8a_worker/graphutils.py
```diff
@@ -10,9 +10,6 @@
                            (host=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_HOST", "localhost"),
                             port=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_PORT", "8182"))
 
-PGM_URL_REST = "http://{host}:{port}".format(host=os.environ.get("PGM_SERVICE_HOST"),
```
msrb: You're still importing `PGM_URL_REST` in `recommender.py`.

f8a_worker/workers/recommender.py
miteshvp: Nice. Thanks.

f8a_worker/workers/recommender.py
miteshvp: We need to handle multiple ecosystem scenario. Currently we handle multiple pom, but we may not be able to handle pom with requirements.txt with this approach

f8a_worker/graphutils.py
miteshvp: Correct, now I create this PGM URL on the fly in recommender direclty

PR #180, opened 2017-08-01
[DigesterTask] compute digests of artifact only

jpopelka:
see https://github.com/fabric8-analytics/fabric8-analytics-worker/issues/175#issuecomment-319297899


msrb:
Good enough workaround for now 😉 Thanks.

PR #183, opened 2017-08-01
Allow to run scancode in more parallel processes.

jpopelka:


msrb:
Hmm, CPU limit for worker container is just 500m. I wonder how will this behave. Let's test it and we will see. Thanks 😉

PR #204, opened 2017-08-03
introduce version filters

miteshvp:


fridex:
Reverting as merge broke CI, will fix in revert.

PR #210, opened 2017-08-04
Multi manifest support v2

jyasveer:


f8a_worker/workers/stackaggregator_v2.py
```diff
@@ -207,16 +206,19 @@ def _get_dependency_data(self, resolved, ecosystem):
...
             resolved = result['details'][0]['_resolved']
             ecosystem = result['details'][0]['ecosystem']
             manifest = result['details'][0]['manifest_file']
+            manifest_file_path = result['details'][0]['manifest_file_path']
 
             finished = self._get_dependency_data(resolved, ecosystem)
-            if finished is not None:
-                stack_data = aggregate_stack_data(finished, manifest, ecosystem.lower(), resolved)
-                return stack_data
+            if finished != None:
```
jpopelka: You [should](https://www.python.org/dev/peps/pep-0008/#programming-recommendations) keep using `is not None` instead of `!= None`.

PR #215, opened 2017-08-04
Incorporate new version of license analysis

harjinder-hari:


f8a_worker/workers/stackaggregator_v2.py
```diff
@@ -117,12 +117,30 @@ def aggregate_stack_data(stack, manifest_file, ecosystem, deps):
...
+    flag_stack_license_exception = False
+    resp = {}
     try:
         license_req = get_session_retry().post(license_url, data=json.dumps(payload))
         resp = license_req.json()
-        stack_license = [resp.get('stack_license')]
     except:
-        stack_license = [None]
+        flag_stack_license_exception = True
+
```
samuzzal-choudhury: Good job! Logic looks good to me. But, can you please move the logic to call license service and subsequent license data population steps in a function?

PR #224, opened 2017-08-08
[DataNormalizer] declared_license (str) -> declared_licenses (list)

jpopelka:
`MercatorTask` (metadata scanner) output contains `declared_license`, which is a string with comma separated license names.

We want it to be a list/array, hence this request.


msrb:
[test this]

PR #233, opened 2017-08-10
Bump selinon and selinonlib versions

pkajaba:
Development of selinon slowed down a little bit so we don't have to rely on git hash anymore. This way we don't have to use update_selinon.sh hacks in server, jobs and firehose-fetcher.


pkajaba:
[test this]

PR #259, opened 2017-08-17
Added new configuration backend - SingleItemBackend

pkajaba:
- Unified accessing of postgres configuration
- Removed env.sh hack


pkajaba:
[test]

PR #265, opened 2017-08-18
Rename cucos-lib-tests to worker-tests

jpopelka:
That's the test image name we [use in runtest.sh](https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/runtest.sh#L16)

Or maybe I don't understand how it works now ?


msrb:
> That's the test image name we use in runtest.sh

Good catch! It was always `cucos-lib-tests`, but the `runtest.sh` script switched to `worker-tests` 2 months ago. The reason why this is still working is because `runtest.sh` builds the `worker-tests` image if it doesn't exist (which means during first worker build every day) and then sources are mounted into the container.

PR #291, opened 2017-08-31
Gather package description from repositories

fridex:


fridex:
Based on: https://github.com/openshiftio/openshift.io/issues/728 and https://github.com/openshiftio/openshift.io/issues/727

PR #293, opened 2017-09-01
Changes to get manifests data from RDS

samuzzal-choudhury:


f8a_worker/workers/graphaggregator.py
```diff
@@ -23,10 +24,20 @@ def _handle_external_deps(self, ecosystem, deps):
 
     def execute(self, arguments):
         self._strict_assert(arguments.get('data'))
+        self._strict_assert(arguments.get('external_request_id'))
+
+        db = self.storage.session
+        results = db.query(StackAnalysisRequest)\
+                        .filter(StackAnalysisRequest.id == arguments.get('external_request_id'))
+        manifests = []
+        if results.count() > 0:
```
jpopelka: With 70c3deec we wrapped all db accessing calls into try-except in order to handle situations when worker gets stuck after pgbouncer restart. Would be nice to have it here (and in `ManifestKeeperTask`) too. Also, you don't have to call `count()` before `first()`, because `first()` returns `None` if the count is 0.

```python
try:
    first = results.first()
except SQLAlchemyError:
    PostgresBase.session.rollback()
    raise
if first is not None:
    row = first.to_dict()
    request_json = row.get("requestJson", {})
    manifests = request_json.get('manifest', [])
```

msrb: @jpopelka I feel like not sharing the session would be a better solution to the reconnect problem - I will comment in the original issue. @samuzzal-choudhury what Jirka suggested is the best way how to go about it for now.

samuzzal-choudhury:
@jpopelka for your review again :-)

samuzzal-choudhury:
Build status: https://ci.centos.org/job/devtools-fabric8-analytics-worker-fabric8-analytics/16/console

Issue #270, opened 2017-08-22
[keywords_tagging]  LookupError(resource_not_found)


jpopelka:
Not sure whether the exception should be catched in worker or tagger:

```
worker-ingestion_1      | Traceback (most recent call last):
worker-ingestion_1      |   File "/usr/lib/python3.4/site-packages/nltk/corpus/util.py", line 80, in __load
worker-ingestion_1      |     try: root = nltk.data.find('{}/{}'.format(self.subdir, zip_name))
worker-ingestion_1      |   File "/usr/lib/python3.4/site-packages/nltk/data.py", line 653, in find
worker-ingestion_1      |     raise LookupError(resource_not_found)
worker-ingestion_1      | LookupError: 
worker-ingestion_1      | **********************************************************************
worker-ingestion_1      |   Resource 'corpora/wordnet.zip/wordnet/' not found.  Please use
worker-ingestion_1      |   the NLTK Downloader to obtain the resource:  >>> nltk.download()
worker-ingestion_1      |   Searched in:
worker-ingestion_1      |     - '/workdir/nltk_data'
worker-ingestion_1      |     - '/usr/share/nltk_data'
worker-ingestion_1      |     - '/usr/local/share/nltk_data'
worker-ingestion_1      |     - '/usr/lib/nltk_data'
worker-ingestion_1      |     - '/usr/local/lib/nltk_data'
worker-ingestion_1      | **********************************************************************
worker-ingestion_1      | 
worker-ingestion_1      | During handling of the above exception, another exception occurred:
worker-ingestion_1      | 
worker-ingestion_1      | Traceback (most recent call last):
worker-ingestion_1      |   File "/usr/lib/python3.4/site-packages/celery/app/trace.py", line 367, in trace_task
worker-ingestion_1      |     R = retval = fun(*args, **kwargs)
worker-ingestion_1      |   File "/usr/lib/python3.4/site-packages/celery/app/trace.py", line 622, in __protected_call__
worker-ingestion_1      |     return self.run(*args, **kwargs)
worker-ingestion_1      |   File "/usr/lib/python3.4/site-packages/selinon/selinonTaskEnvelope.py", line 170, in run
worker-ingestion_1      |     raise self.retry(max_retries=0, exc=exc)
worker-ingestion_1      |   File "/usr/lib/python3.4/site-packages/celery/app/task.py", line 668, in retry
worker-ingestion_1      |     raise_with_context(exc)
[...]

PR #297, opened 2017-09-05
Run prepare for tagger

fridex:


fridex:
Please wait for @pkajaba's changes in tagger before merging. I will modify git hash accordingly afterwards.

fridex:
OK, so let's proceed for now with this one, we can add Maven collector later.

Dockerfile
```diff
@@ -43,3 +43,6 @@ RUN find ${HOME} -mindepth 1 -delete
 RUN mkdir -p /tmp/install_deps/patches/
 COPY hack/patches/* /tmp/install_deps/patches/
 RUN /tmp/install_deps/patches/apply_patches.sh
+
+# Install external resources needed for automated tagging
+RUN python3 -c 'import f8a_tagger; f8a_tagger.prepare()'
```
jpopelka: Could we move this closer to the tagger installation, i.e. line 37 ?

fridex:
Fixes: https://github.com/fabric8-analytics/fabric8-analytics-worker/issues/270

Dockerfile
fridex: Sure, will do.

fridex: Or we could add this directly to the install script.

PR #299, opened 2017-09-05
Rename stack analysis flow

fridex:


f8a_worker/dispatcher/flows/stackApiGraphFlow.yml
```diff
@@ -1,14 +1,19 @@
 ---
   flow-definitions:
-    - name: 'stackApiGraphFlow'
+    - name: 'stackApiGraphV2Flow'
```
jpopelka: Why ? From the rest of the change it looks like we're removing `stackApiGraphV2Flow`

fridex: Good catch!

fridex:
I will close this and leave possible renaming on @samuzzal-choudhury as this change needs renaming in server repo as well.

PR #311, opened 2017-09-08
Introduce priority worker

fridex:


msrb:
Just that we are on the same page:
- Pavel's firehose-fetcher will talk to ingestion queues (as it basically schedules new analyses)
- data refreshes will happen in these new refresh queues
- when system is running as expected, all (api, ingestion, refresh) queues will be occupied

Is this what we want here?

fridex:
> Is this what we want here?

We have basically 3 scenarios for workers:

* requests sent from API server to prioritize user requests (SLA)
  * api worker queues
* requests sent based on our needs (mostly manually triggered) - initial ingest of new ecosystem, introduction of new tasks (old results recomputed), bugfixes in old tasks (recompute stored results), ...
  * ingestion queue
* requests sent based on new releases (firehose-fetcher for now)
  * this would be done on a separate queue not to affect our manually triggered analyses
  * this was the desired scenario for a new worker type - so something like `updates`/`releases` would be more accurate naming?

Note that this is only initial worker introduction, I will need to change dispatcher config files accordingly (based on agreed naming).

msrb:
The real question is: what is the problem we are trying to solve here? Should we be worried about firehose-fetcher overloading the system? (cc @pkajaba ) Numbers in http://www.modulecounts.com/ don't look that scary, in my opinion.

What would be the queue for the real data refreshes, like for example GitHub data refresh?

If the real problem here is that all queues are occupied and therefore it's hard to proceed with work like tagging, then I am not sure if dedicated queue for firehose-fetcher will do the trick here.

So it really depends on what's the real problem here 😉

jpopelka:
> requests sent based on new releases (firehose-fetcher for now)
>     so something like updates/releases would be more accurate naming?

If it's only for the new releases, then `releases` is better than `refresh` indeed, because `refresh` sound more like 'recompute old results' to me.

fridex:
To document the outcome that we agreed on:

We will introduce one new worker type - "`priority` workers" which would be used for manual interaction with the system if there is a need (e.g. run a certain task, trigger rerun of some tasks, ingestion of new ecosystem).

This worker type will be used in parallel to API workers (SLA on API side) and in parallel to `ingestion` workers that will be used for periodic jobs triggered by Jobs service (e.g. refresh of GitHub data, CVE flagging) or for analysing new package releases (firehose fetcher).

I will update PRs accordingly. Thanks!

fridex:
Relevant PRs:

 * https://github.com/fabric8-analytics/fabric8-analytics-common/pull/272
 * https://github.com/openshiftio/saas-analytics/pull/83
 * https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/311

PR #316, opened 2017-09-11
Fix broken tests

fridex:


fridex:
This is obvious fix, merging myself. Sorry for that :)

PR #317, opened 2017-09-11
Do proper fix on tests

fridex:


fridex:
I need this in production, I will reiterate the proper fix.

PR #328, opened 2017-09-13
Compute keywords only if there is some input

fridex:


fridex:
> LGMT, maybe one suggestion. Those 3 blocks of code are more or less same, so you can create a util function and parametrize it.

I'm 50:50 about this - it's 3 lines of code. I will merge as this is needed in production. Will try to make with a more nice solution then. :)

PR #335, opened 2017-09-13
[WIP] Dataimporter as worker

miteshvp:


miteshvp:
@fridex - PTAL

miteshvp:
related local development changes - https://github.com/fabric8-analytics/fabric8-analytics-common/pull/288

hack/create_graph_schema.py
```diff
@@ -0,0 +1,52 @@
...
+while not SCHEMA_CREATED:
+    logger.info("Creating Graph Schema Now...")
+    print("Creating Graph Schema Now...")
+    payload = {'gremlin': str_gremlin_dsl}
+
+    response = requests.post(GREMLIN_SERVER_URL_REST,
+                             data=json.dumps(payload))
+    json_response = response.json()
+
+    if response.status_code != 200:
```
fridex: What happens if there will be done schema creation like this multiple times? Will it fail? If yes, the container will keep restarting in production.

hack/create_graph_schema.py
```diff
@@ -0,0 +1,52 @@
...
+        logger.info("ERROR %d(%s): %s" % (response.status_code, response.reason, json_response.get("message")))
+        logger.info("ERROR Creating Schema. Waiting for 10 Seconds")
+        print("ERROR Creating Schema. Waiting for 10 Seconds")
```
fridex: Decide whether you want to use logger or just prints here. Don't do both.

hack/create_graph_schema.py
```diff
@@ -0,0 +1,52 @@
...
+        time.sleep(10)
```
fridex: This should be done by OpenShift. Fail with error so OpenShift reschedules schema creation transparently.

f8a_worker/workers/graph_importer_utils.py
```diff
@@ -0,0 +1,204 @@
...
+    return obj
+
+
+def _first_key_info(first_key, bucket_type):
+    obj = {}
+    t = read_json_file(first_key, bucket_type)
+    obj["dependents_count"] = t.get("dependents_count", '-1')
+    obj["package_info"] = t.get("package_info", {})
+    obj["latest_version"] = t.get("latest_version", '-1')
+    return obj
```
fridex: Just a question: why we do not sync data as we have them on S3 without adding additional data to them? Syncs would be more transparent and lightweight, issues will be easier to track.

f8a_worker/workers/graph_importer_utils.py
```diff
@@ -0,0 +1,204 @@
...
+
+def _import_keys_from_s3_http(epv_list):
+    logger.debug("Begin import...")
+    report = {'status': 'Success', 'message': 'The import finished successfully!'}
+    count_imported_EPVs = 0
+    last_imported_EPV = None
+    epv = []
+    for epv_key in epv_list:
+        for key, contents in epv_key.items():
+            if len(contents.get('pkg_list_keys')) == 0 and len(contents.get('ver_list_keys')) == 0:
```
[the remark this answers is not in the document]
miteshvp: Yes. If selective_tasks is an empty array, it will do all the syncs. You can find it in line#169 and line#174 of this file.

f8a_worker/workers/graph_importer_utils.py
```diff
@@ -0,0 +1,204 @@
...
+logger = logging.getLogger(__name__)
+
+# Get Variable Parameters from Environment
+GREMLIN_SERVER_URL_REST = "http://{host}:{port}".format(
+                           host=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_HOST", "localhost"),
+                           port=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_PORT", "8182"))
+s3 = s3_pkg = None
+
+
+def read_json_file(filename, bucket_type):
```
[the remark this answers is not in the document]
miteshvp: It uses `retrieve_dict()` in this function

f8a_worker/storages/s3.py
```diff
@@ -195,6 +195,16 @@ def retrieve_dict(self, object_key):
         """ Retrieve a dictionary stored as JSON from S3 """
         return json.loads(self.retrieve_blob(object_key).decode())
 
+    def retrieve_key_list(self, prefix=None):
+        """ Retrieve all the keys in S3 based on a prefix"""
```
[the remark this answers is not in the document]
miteshvp: This is needed in case we want to do force_graph_syncs. We can just do `selective_tasks = []` to sync all the data. I don't know, but I am assuming, we can speed up things with this approach since we can directly invoke GraphImporter task and still do the syncs directly without invoking any other un-necessary task. Does it make sense?

openshift/template.yaml
```diff
@@ -24,6 +24,14 @@ objects:
...
             containerName: bayesian-worker
             command: ["/alembic/run-db-migrations.sh"]
+    strategy:
+      type: Rolling
+      rollingParams:
+        pre:
+          failurePolicy: Retry
+          execNewPod:
+            containerName: bayesian-worker
+            command: ["python3 /graph/create_graph_schema.py"]
```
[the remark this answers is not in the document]
miteshvp: ok. done.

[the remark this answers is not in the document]
miteshvp: done. thanks

f8a_worker/workers/graph_importer.py
```diff
@@ -22,30 +13,21 @@ def execute(self, arguments):
...
+                    'epv': package_list,
+                    'count_imported_EPVs': report.get('count_imported_EPVs')}
 
-        if response.status_code != 200:
-            raise RuntimeError("Failed to invoke graph import at '%s' for %s" % (endpoint, param))
+        self.log.debug(response)
 
-        self.log.info("Graph import succeeded with response: %s", response.text)
+        if report.get('status') is not 'Success':
+            raise RuntimeError("Failed to ingest data for '%s' with %s" % (package_list, select_ingest))
```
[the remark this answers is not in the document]
miteshvp: agreed. I guess this is redundant in a way we are logging the message. Done.

f8a_worker/workers/graph_importer_utils.py
```diff
@@ -0,0 +1,204 @@
...
+from selinon import StoragePool
+
+
+logger = logging.getLogger(__name__)
+
+# Get Variable Parameters from Environment
+GREMLIN_SERVER_URL_REST = "http://{host}:{port}".format(
+                           host=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_HOST", "localhost"),
+                           port=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_PORT", "8182"))
+s3 = s3_pkg = None
```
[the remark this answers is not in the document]
miteshvp: Yup. Agreed. I am using S3 Data storage. Done the changes.

f8a_worker/workers/graph_importer_utils.py
miteshvp: I am not following your question here. Is it, why are we creating a dict here instead of using it directly elsewhere?

f8a_worker/workers/graph_importer_utils.py
```diff
@@ -0,0 +1,204 @@
...
+
+    return report
+
+
+def import_epv_from_s3_http(list_epv, select_doc=None):
+
+    try:
+        global s3, s3_pkg
+        s3 = StoragePool.get_connected_storage('S3Data')
+        s3_pkg = StoragePool.get_connected_storage('S3PackageData')
```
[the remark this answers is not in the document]
miteshvp: done.

f8a_worker/workers/graph_importer_utils.py
```diff
@@ -0,0 +1,204 @@
...
+        # Collect relevant files from data-source and group them by package-version.
+        list_keys = []
+        for epv in list_epv:
+            dict_keys = {}
+            ver_list_keys = []
+            pkg_list_keys = []
+            pkg_key_prefix = ver_key_prefix = ""
+
+            if 'name' not in epv or 'ecosystem' not in epv:
+                continue
```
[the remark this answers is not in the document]
miteshvp: yes it is only one EPV from worker. But, I would still prefer to let it be, just in future, we may want to bypass all except graph syncs.

f8a_worker/workers/graph_importer_utils.py
```diff
@@ -0,0 +1,204 @@
...
+            elif 'version' not in epv or epv.get('version') is None:
+                epv['version'] = ''
```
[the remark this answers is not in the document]
miteshvp: This is more to do with graph syncs and anything to do with workers. This will help us to sync multiple EPVs (some for package, some for pacakge-version). This may be helpful to invoke standalone graphimporter workflow. I would rather keep it unless it really causes some issues that I might not able to follow.

f8a_worker/workers/graph_populator.py
```diff
@@ -0,0 +1,231 @@
...
+
+        drop_props = []
+        str_version = prp_version = drop_prop = ""
+        # Check if license and cve analyses succeeded. Then we refresh the property
+        if 'success' == input_json.get('analyses', {}).get('source_licenses', {}).get('status'):
+            drop_props.append('licenses')
+        if 'success' == input_json.get('analyses', {}).get('security_issues', {}).get('status'):
+            drop_props.append('cve_ids')
+        if len(drop_props) > 0:
+            drop_prop += "g.V().has('pecosystem','" + ecosystem + "').has('pname','" + pkg_name + "')" \
```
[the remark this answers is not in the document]
miteshvp: Yes. Totally agreed. I will create another issue to keep a track of this. WDYT?

f8a_worker/workers/graph_populator.py
```diff
@@ -0,0 +1,231 @@
...
+
+        return str_version
+
+    @classmethod
+    def construct_package_query(cls, input_json):
+        pkg_name = input_json.get('package')
+        ecosystem = input_json.get('ecosystem')
+        pkg_name_tokens = re.split('\W+', pkg_name)
+        prp_package = ""
+        str_package = "pkg = g.V().has('ecosystem','" + ecosystem + "').has('name','" + pkg_name + "').tryNext()" \
```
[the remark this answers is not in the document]
miteshvp: +1

f8a_worker/workers/graph_populator.py
```diff
@@ -0,0 +1,231 @@
...
+
+        # Add CVE property if it exists
+        if 'security_issues' in input_json.get('analyses', {}):
+            cves = []
+            for cve in input_json.get('analyses', {}).get('security_issues', {}).get('details', []):
+                cves.append(cve.get('id') + ":" + str(cve.get('cvss', {}).get('score')))
+            prp_version += " ".join(map(lambda x: "ver.property('cve_ids', '" + x + "');", cves))
+
+        # Get Metadata Details
+        if 'metadata' in input_json.get('analyses', {}):
```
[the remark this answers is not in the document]
miteshvp: I guess this was more a personal preference. Made it easier for me to keep track on graph query being generated.

hack/create_graph_schema.py
```diff
@@ -0,0 +1,52 @@
...
+with open(schema_file_path, 'r') as f:
+    str_gremlin_dsl = f.read()
+
+while not SCHEMA_CREATED:
+    logger.info("Creating Graph Schema Now...")
+    print("Creating Graph Schema Now...")
+    payload = {'gremlin': str_gremlin_dsl}
+
+    response = requests.post(GREMLIN_SERVER_URL_REST,
+                             data=json.dumps(payload))
```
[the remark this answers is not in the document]
miteshvp: Sure.

hack/create_graph_schema.py
miteshvp: Nope. The way schema creation is done, it will create schema if not present, or it will create additional schema items recently added, or it will just pass. All of it returns a success code. Did I clarify your doubt?

hack/create_graph_schema.py
miteshvp: I just want to inform the developer to know what is happening while graph_schema creation. Print should just do fine, I guess. Removed logger statements. Thanks.

hack/create_graph_schema.py
miteshvp: done.

f8a_worker/workers/graph_importer_utils.py
```diff
@@ -0,0 +1,204 @@
...
+    obj = {"analyses": {}}
+    for this_key in other_keys:
+        value = read_json_file(this_key, bucket_type)
+        this_key = this_key.split("/")[-1]
+        if 'success' == value.get('status', ''):
+            obj["analyses"][this_key[:-len('.json')]] = value
+    return obj
+
+
+def _get_exception_msg(prefix, e):
```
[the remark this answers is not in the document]
miteshvp: Added exception message in this function

f8a_worker/workers/graph_importer_utils.py
```diff
@@ -0,0 +1,204 @@
...
+                    resp = response.json()
+
+                    if 'status' in resp and resp['status']['code'] == 200:
+                        count_imported_EPVs += 1
+                        last_imported_EPV = obj.get('ecosystem') + ":" + obj.get('package') + ":" + obj.get('version')
+                    elif 'Exception-Class' in resp:
+                        report['message'] = "The import failed " + resp['message']
+                        report['status'] = 'Failure'
+
+            except Exception as e:
```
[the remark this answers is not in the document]
miteshvp: Added logger.exception message

f8a_worker/storages/s3.py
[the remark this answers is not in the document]
fridex: > This is needed in case we want to do force_graph_syncs. We can just do selective_tasks = [] to sync all the data. I don't know, but I am assuming, we can speed up things with this approach since we can directly invoke GraphImporter task and still do the syncs directly without invoking any other un-necessary task. Does it make sense?

This is handled by our orchestrator - Selinon/Dispatcher that computes which tasks need to be run and if there are any that need to be run. In case you want to get all tasks for the given package, please create such method in S3 adapter. In this case there should be 2 methods - one in `S3Data` to handle package-version level data and another in `S3PackageData` to handle package-level data. `S3Data`'s method should accept ecosystem, package, version parameters and `S3PackageData`'s method should accept ecosystem package parameters. The logic should be encapsulated so callee doesn't need to know internal structure of how data are stored - so no prefix argument should be passed.

f8a_worker/workers/graph_importer_utils.py
fridex: > I am not following your question here. Is it, why are we creating a dict here instead of using it directly elsewhere?

Kind of. What is the intention behind this function?

f8a_worker/workers/graph_importer_utils.py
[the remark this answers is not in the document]
fridex: Actually I was referring to removing this function whatsoever. If there is some error, just raise appropriate exception and leave exception handling on higher levels if needed. Getting exception message is not necessary in this code - we will transparently aggregate, report and format them in appropriate layers.

f8a_worker/workers/graph_importer_utils.py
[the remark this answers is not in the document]
fridex: I would rather keep code minimal, easy to maintain, easy to extend and easy to read without adding unnecessary complexity that could easily lead to errors. We can still get back to the previous version based on git logs (see recent "reintroduction" of go support). WDYT?

f8a_worker/workers/graph_importer_utils.py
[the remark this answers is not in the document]
fridex: The same as above - I would rather keep code minimal and doing desired actions properly instead of introducing unnecessary bulking complexity.

f8a_worker/workers/graph_importer_utils.py
miteshvp: This is more about creating the dict that graph sync understands. In this case, all the relevant meta data information about the package to be synced to graph resides in something like `<ecosystem>/<package>/<version>.json`. We just use this to modify graph sync object accordingly.

f8a_worker/workers/graph_importer_utils.py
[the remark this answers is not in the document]
miteshvp: I personally feel we should leave this functionality intact at least for the time being, this gives us flexibility in the future to just invoke graph syncs by-passing all other tasks.. Let's ingest one ecosystem with graph sync and check whether this would make any sense.

f8a_worker/workers/graph_importer_utils.py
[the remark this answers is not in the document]
miteshvp: Done.

f8a_worker/storages/s3.py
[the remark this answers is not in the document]
miteshvp: Ok, i see what you mean. Done the necessary changes. Let me know if you are ok

hack/create_graph_schema.py
```diff
@@ -0,0 +1,46 @@
...
+GREMLIN_SERVER_URL_REST = "http://{host}:{port}".format(
+                          host=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_HOST", "localhost"),
+                          port=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_PORT", "8182"))
+
+GRAPH_INITIALIZED = False
+SCHEMA_CREATED = False
+# wait for graph to be initialized
+while not GRAPH_INITIALIZED:
+    try:
+        print("Calling GREMLIN - " + GREMLIN_SERVER_URL_REST)
```
pkajaba: please use logging here instead of print

hack/create_graph_schema.py
```diff
@@ -0,0 +1,46 @@
...
+        resp = requests.get(GREMLIN_SERVER_URL_REST)
+        if resp.status_code >= 200:
```
pkajaba: `resp.ok` might be better here. I leave it up to you.

hack/create_graph_schema.py
```diff
@@ -0,0 +1,46 @@
...
+        continue
+
+# send the schema payload for creation
+current_file_path = os.path.dirname(os.path.realpath(__file__))
+schema_file_path = os.path.join(current_file_path, 'schema.groovy')
+with open(schema_file_path, 'r') as f:
+    str_gremlin_dsl = f.read()
+
+while not SCHEMA_CREATED:
+    print("Creating Graph Schema Now...")
```
pkajaba: please you `logger` here again instead of print

hack/create_graph_schema.py
```diff
@@ -0,0 +1,46 @@
...
+    payload = {'gremlin': str_gremlin_dsl}
```
pkajaba: you don't have to define `payload` variable here just move it to `requests.post`

hack/create_graph_schema.py
miteshvp: Tried with logging, but while tailing the logs, it will print to the console only when it exits. Print looks suitable here. Again, this is just for user's information when they want to tail the logs.

hack/create_graph_schema.py
```diff
@@ -0,0 +1,46 @@
...
+
+    response = requests.post(GREMLIN_SERVER_URL_REST, json=payload)
+    json_response = response.json()
+
+    if response.status_code != 200:
```
pkajaba: I wonder what happened here if gremlin returns `201`. It fails with `RuntimeError`, but will it be consistent when we want to create same scheme again even though it was successful last time?

hack/create_graph_schema.py
miteshvp: same explanation as above. Tailing logs was not giving expected results

hack/create_graph_schema.py
miteshvp: Yes. If schema is not created, it will create afresh.  Even if schema is created in the first place, it will just pass through but still send 200. Plus, if its not 200, then gremlin rolls back the entire transaction.

hack/create_graph_schema.py
```diff
@@ -0,0 +1,46 @@
...
+schema_file_path = os.path.join(current_file_path, 'schema.groovy')
+with open(schema_file_path, 'r') as f:
+    str_gremlin_dsl = f.read()
+
+while not SCHEMA_CREATED:
+    print("Creating Graph Schema Now...")
+    payload = {'gremlin': str_gremlin_dsl}
+
+    response = requests.post(GREMLIN_SERVER_URL_REST, json=payload)
+    json_response = response.json()
```
pkajaba: you are not really using this variable so you can call `response.json().get("message")` in logs

hack/create_graph_schema.py
miteshvp: better readability

hack/create_graph_schema.py
miteshvp: just readability sake

f8a_worker/workers/graph_populator.py
```diff
@@ -0,0 +1,242 @@
+import logging
+import re
+import time
+from datetime import datetime
+
+logger = logging.getLogger(__name__)
+
+# Compile patterns to clean a string (used for safety against GraphQL injection attack)
+danger_word_list = ["drop\(\)", "V\(\)", "count\(\)"]
```
pkajaba: some test for these patters would be nice

miteshvp: good suggestion

f8a_worker/workers/graph_populator.py
```diff
@@ -0,0 +1,242 @@
...
+
+class GraphPopulator(object):
+    @classmethod
+    def str_value_cleaner(cls, safe_string):
+        """
+        Clean up the safe_string to remove suspicious words,
+        and special characters.
+        """
+        result = ''
+        if isinstance(safe_string, str):  # can raise an exception as well
```
pkajaba: what does `can raise an exception as well` means?

f8a_worker/workers/graph_populator.py
```diff
@@ -0,0 +1,242 @@
...
+
+    @classmethod
+    def construct_version_query(cls, input_json):
+        description = ''
+        try:
+            if len(input_json.get('analyses', {}).get('metadata', {}).get('details')) > 0:
+                description = input_json.get('analyses').get('metadata').get('details')[0].get('description', '')
+                description = cls.str_value_cleaner(description).replace("'", "\\'")
+        except:
+            # we pass and move forward without description
```
pkajaba: maybe debug log here would be nice that we skipped description.

f8a_worker/workers/graph_populator.py
```diff
@@ -0,0 +1,242 @@
...
+            for lang in input_json.get('analyses').get('code_metrics').get('details', {}).get('languages', []):
+                if lang.get('metrics', {}).get('functions', {}).get('average_cyclomatic_complexity'):
+                    count += 1
+                    tot_complexity += lang['metrics']['functions']['average_cyclomatic_complexity']
+            cm_avg_cyclomatic_complexity = str(tot_complexity / count) if count > 0 else '-1'
+            cm_loc = str(input_json.get('analyses').get('code_metrics').get('summary', {}).get('total_lines', -1))
+            cm_num_files = str(input_json.get('analyses').get('code_metrics').get('summary', {})
+                               .get('total_files', -1))
+            prp_version += "ver.property('cm_num_files'," + cls.str_value_cleaner(cm_num_files) + ");" \
+                           "ver.property('cm_avg_cyclomatic_complexity'," + cls.str_value_cleaner(cm_avg_cyclomatic_complexity) + ");" \
```
pkajaba: This line has too many symbols.

f8a_worker/workers/graph_populator.py
```diff
@@ -0,0 +1,242 @@
...
+        str_version = prp_version = drop_prop = ""
+        # Check if license and cve analyses succeeded. Then we refresh the property
+        if 'success' == input_json.get('analyses', {}).get('source_licenses', {}).get('status'):
+            drop_props.append('licenses')
+        if 'success' == input_json.get('analyses', {}).get('security_issues', {}).get('status'):
+            drop_props.append('cve_ids')
+        if len(drop_props) > 0:
+            drop_prop += "g.V().has('pecosystem', ecosystem).has('pname', pkg_name)" \
+                         ".has('version', version).properties('" + "','".join(drop_props) + "')." \
+                         "drop().iterate();" \
```
pkajaba: what is a purpose of this backslash?

f8a_worker/workers/graph_populator.py
```diff
@@ -0,0 +1,242 @@
...
+        str_version += "ver = g.V().has('pecosystem', ecosystem).has('pname', pkg_name)" \
+                       ".has('version', version).tryNext().orElseGet{graph.addVertex('pecosystem', ecosystem," \
+                       "'pname', pkg_name, 'version', version, " \
+                       "'vertex_label', 'Version')};" \
+                       "ver.property('last_updated'," + str(time.time()) + ");"
+        # Add Description if not blank
+        if description:
+            prp_version += "ver.property('description','" + re.sub('[^A-Za-z0-9_\\\/\'":. ]', '', description) + "');"
+        # Get Code Metrics Details
+        if 'code_metrics' in input_json.get('analyses', {}):
```
pkajaba: What if there is no `code_metrics`. Is it correct state?

f8a_worker/workers/graph_populator.py
```diff
@@ -0,0 +1,242 @@
...
+            prp_package += "pkg.property('latest_version', '" + latest_version + "');"
+
+        # Get Github Details
+        if 'github_details' in input_json.get('analyses', {}):
+            gh_details = input_json.get('analyses').get('github_details').get('details', {})
+            gh_prs_last_year_opened = str(gh_details.get('updated_pull_requests', {}).get('year', {}).get('opened', -1))
+            gh_prs_last_month_opened = str(gh_details.get('updated_pull_requests', {}).get('month', {}).get('opened', -1))
+            gh_prs_last_year_closed = str(gh_details.get('updated_pull_requests', {}).get('year', {}).get('closed', -1))
+            gh_prs_last_month_closed = str(gh_details.get('updated_pull_requests', {}).get('month', {}).get('closed', -1))
+            gh_issues_last_year_opened = str(gh_details.get('updated_issues', {}).get('year', {}).get('opened', -1))
```
pkajaba: Why is default value here '-1'? I know that it was here before, but what is a reasoning behind this value.

f8a_worker/workers/graph_populator.py
```diff
@@ -0,0 +1,242 @@
...
+            gh_issues_last_month_opened = str(gh_details.get('updated_issues', {}).get('month', {}).get('opened', -1))
+            gh_issues_last_year_closed = str(gh_details.get('updated_issues', {}).get('year', {}).get('closed', -1))
+            gh_issues_last_month_closed = str(gh_details.get('updated_issues', {}).get('month', {}).get('closed', -1))
+            gh_forks = str(gh_details.get('forks_count', -1))
+            gh_stargazers = str(gh_details.get('stargazers_count', -1))
+            gh_open_issues_count = str(gh_details.get('open_issues_count', -1))
+            gh_subscribers_count = str(gh_details.get('subscribers_count', -1))
+
+            prp_package += "pkg.property('gh_prs_last_year_opened', " + cls.str_value_cleaner(gh_prs_last_year_opened) + ");" \
```
pkajaba: would not format method work here nicer?

f8a_worker/workers/graph_populator.py
```diff
@@ -0,0 +1,242 @@
...
+                                             .get('releases', {}).get('count', -1)
+            libio_latest_release = input_json.get('analyses').get('libraries_io').get('details', {}) \
+                                             .get('releases', {}).get('latest', {}).get('published_at')
+            libio_latest_version = input_json.get('analyses').get('libraries_io').get('details', {}) \
+                                             .get('releases', {}).get('latest', {}).get('version', '')
+
+            if libio_latest_release is not None:
+                try:
+                    prp_package += "pkg.property('libio_latest_release', '" + \
+                                   str(time.mktime(datetime.strptime(libio_latest_release,
```
pkajaba: Why is this date magic is required here?

hack/create_graph_schema.py
```diff
@@ -0,0 +1,46 @@
+import os
+import requests
+import json
+import time
+import logging
+
+logger = logging.getLogger(__name__)
+
+GREMLIN_SERVER_URL_REST = "http://{host}:{port}".format(
```
pkajaba: you can import this from graph_utils

f8a_worker/workers/graph_importer_utils.py
```diff
@@ -0,0 +1,183 @@
...
+
+logger = logging.getLogger(__name__)
+
+# Get Variable Parameters from Environment
+GREMLIN_SERVER_URL_REST = "http://{host}:{port}".format(
+                           host=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_HOST", "localhost"),
+                           port=os.environ.get("BAYESIAN_GREMLIN_HTTP_SERVICE_PORT", "8182"))
+
+
+class GETS3:
```
pkajaba: This class is sort of useless, why is it here?

pkajaba:
In general this good, but the way we work with `gremlin` is just bad. We have to introduce some kind of library/layer which will translate gremlin calls since this concentration of strings might be dangerous in future.

Also, I would like to see some tests for this service.

hack/create_graph_schema.py
fridex: > same explanation as above. Tailing logs was not giving expected results

How did you configure logging facilities?

hack/create_graph_schema.py
pkajaba: Can you please elaborate how users are supposed to take advantage of print here? It acts the same way as print it's just unified. Your problem might be with a) setup of logging here b) level of logging.

hack/create_graph_schema.py
miteshvp: tried logger.debug and logger.info

f8a_worker/workers/graph_populator.py
miteshvp: maybe a leftover.

f8a_worker/workers/graph_populator.py
miteshvp: if there is no code_metrics collected it will just skip

f8a_worker/workers/graph_populator.py
miteshvp: just a way to identify that we do not have corresponding github metrics collected

f8a_worker/workers/graph_populator.py
miteshvp: converts the date into required format

f8a_worker/workers/graph_importer_utils.py
miteshvp: it is used

f8a_worker/workers/graph_populator.py
pkajaba: so it converts date from some format to format which is defined in a schema of the graph. Is that correct?

f8a_worker/workers/graph_populator.py
pkajaba: I figured this out, but I am asking what is it -1.

f8a_worker/workers/graph_populator.py
pkajaba: Is it not missing in the graph then?

f8a_worker/workers/graph_populator.py
pkajaba: :+1:

hack/create_graph_schema.py
fridex: I was speaking about logging configuration - you need to explicitly configure logging (it's silent tby default - see logging levels). you can find more info here https://docs.python.org/3/library/logging.html#logging.basicConfig

hack/create_graph_schema.py
pkajaba: It does not help readability... at all. You are defining a variable which is not really used and if it is it's just in logs.

hack/create_graph_schema.py
pkajaba: sounds good to me. This sounds like a good test case to me.

miteshvp:
> but the way we work with gremlin is just bad. We have to introduce some kind of library/layer which will translate gremlin calls since this concentration of strings might be dangerous in future.
1. Only input that comes from outside world is E:P:V. We have already implemented bindings for the same.
2. Additional safety for any data that is read from S3 buckets, is passed through string cleaning to get rid of danger words in terms of GraphQL.
3. There is no easy way to have a generic library that works with different task results. 
4. Gremlin-HTTP calls is all about post requests with a payload. So to me this would be an un-necessary layer. But I may be wrong.

> Also, I would like to see some tests for this service.
Yes, that would be something @fridex or @pkajaba can help me with. Still getting my head around how tests are designed.

hack/create_graph_schema.py
pkajaba: It depends on point of view, but here it's more justified.

f8a_worker/workers/graph_importer_utils.py
pkajaba: yes, but like an object, this class does not make much sense to me.

f8a_worker/workers/graph_populator.py
miteshvp: Not following you. Line greater than 120 characters? I have pushed the changes, but I just kept it on purpose for better readability.

hack/create_graph_schema.py
miteshvp: It should be same as worker-api/worker-ingestion since its a part of worker. So logger should work by default.

f8a_worker/workers/graph_populator.py
miteshvp: nope. Graph does not require code_metrics as mandatory. Infact, now we do not show code_metrics at all, and we have also disabled in the workflow. @fridex can correct me if I am wrong.

f8a_worker/workers/graph_populator.py
miteshvp: All those github metrics are Integer in graph. So `-1` works on behalf of `null` .

f8a_worker/workers/graph_populator.py
miteshvp: 👍

f8a_worker/workers/graph_importer_utils.py
```diff
@@ -0,0 +1,183 @@
...
+def _first_key_info(first_key, bucket_type):
+    obj = {}
+    t = read_json_file(first_key, bucket_type)
+    obj["dependents_count"] = t.get("dependents_count", '-1')
+    obj["package_info"] = t.get("package_info", {})
+    obj["latest_version"] = t.get("latest_version", '-1')
+    return obj
+
+
+def _other_key_info(other_keys, bucket_type):
```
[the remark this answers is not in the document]
miteshvp: to differentiate between meta json and other task jsons

f8a_worker/workers/graph_populator.py
fridex: [Code metrics is disabled](https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/f8a_worker/dispatcher/flows/bayesianAnalysisFlow.yml#L45). In that case please remove this dead code.

hack/create_graph_schema.py
fridex: > It should be same as worker-api/worker-ingestion since its a part of worker. So logger.debug should have worked by default.

You are creating a standalone process, it has nothing to do with worker code whatsoever.

f8a_worker/workers/graph_importer_utils.py
miteshvp: Is there a better way in your mind? Maybe I am not aware on how to get this handle in a better way.

f8a_worker/workers/graph_populator.py
miteshvp: I feel not the need for.

f8a_worker/workers/graph_importer_utils.py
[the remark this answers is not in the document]
miteshvp: tend to disagree here as I fail to understand what part of it is becoming too complex.

hack/create_graph_schema.py
miteshvp: this code is part of worker and should use same logging configuration as `worker`

f8a_worker/workers/graph_importer_utils.py
fridex: > Is there a better way in your mind? Maybe I am not aware on how to get this handle in a better way.

Just request these adapters in the code where you need them. You don't need to care about these much - JFYI they are connecting lazily and the connection is shared across the whole worker codebase to optimize access (part of worker "global" context). For a user using such adapter - just ask for it and use it's API.

Ideally we would like to introduce the same adapter for graph database as well so we can share implementation in other parts of project (e.g. server, ...). This will minimize code, simplify it and in case of bugs, there will be a straightforward fix at one place.

hack/create_graph_schema.py
miteshvp: ok. So, then i will change the logging configuration. I was under the impression, it will pick same configs as worker.

hack/create_graph_schema.py
miteshvp: got clarification from Frido.

fridex: How would you like to accomplish such things? It's a simple separate process. If you run a web browser besides worker code on your local setup, does it mean that the browser should report errors the same way how worker does?

hack/create_graph_schema.py
miteshvp: Done.

hack/create_graph_schema.py
miteshvp: Done.

hack/create_graph_schema.py
miteshvp: Done.

f8a_worker/workers/graph_populator.py
miteshvp: What are the chances that it will never be enabled?

f8a_worker/workers/graph_importer_utils.py
miteshvp: This looks like outside the scope for this PR. Please go ahead and create a technical debt around this.

f8a_worker/workers/graph_populator.py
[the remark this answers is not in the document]
miteshvp: Handled now

hack/create_graph_schema.py
miteshvp: No special treatment needed. Just want to print to console for any user to follow in local deployment, about whether graph_schema_creation is done or not.

f8a_worker/workers/graph_importer_utils.py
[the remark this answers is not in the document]
pkajaba: so why `_first_key_info` is not named `_meta_key_info`?

f8a_worker/workers/graph_populator.py
pkajaba: Readability is really questionable and subjective. However, our goal is to enable linter. And if all new code will be linter ready we won't have so many issues to enable it if we follow this strictly.

So all we have to do is agree on line length. PEP8 dictates 79 chars.

fridex: > So all we have to do is agree on line length. PEP8 dictates 79 chars.

79 chars is really low. I would suggest at least 100 (this is configurable in linters). We don't have old terminals :)

f8a_worker/workers/graph_importer_utils.py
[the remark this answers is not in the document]
pkajaba: You have way too many conditions, that's why @fridex find code too complex. Sometimes it's really necessary but you can think of an easier way it's always a better option.

I would guess that cyclomatic complexity of this part is unnecessarily high, but I am not sure.

pkajaba:
> There is no easy way to have a generic library that works with different task results.
I am not saying it would be an easy task to create such approach. However, It is definitely worth it. In fact, a naive implementation would not be so hard. You have limited number of entities, so you could create classes which would encapsulate this and query creation would be inside of these classes. 

This way you could spare code duplication... this is especially helpful if you find some mistake/vulnerability because you don't have to look for every occurrence, but you fix just this "serialization" part.

miteshvp:
@fridex @pkajaba - pushed code with graph adapter. This adapter can be built with more functions based on the needs. Currently it serves the purpose for GraphImporter. But for others, this is out of scope for this PR.

pkajaba:
I am not sure if I follow your idea of the adapter. My idea was creating class let's say `Query`. This class would have methods called `addVertex` and add property. Inside these methods, you could concatenate strings and run a check for injections. You could extend this class to `PackageQuery` where you could check for precise syntax/required attributes of a query. In the end, you would serialize and throw an exception if you had some required keys missing. I find something like more robust. Note that this was just ideas and implementation would not be so easy and straightforward.

I don't agree that this is out of the scope of this PR. DataImporter as a worker is supposed to reduce technical debt. It's good to move DataImporter functionality from service to worker, but this is architectural standpoint. Real technical debt is hidden in implementation itself and if we proceed with this and merge, it's not about technical debt reduction, but more about moving technichal debt.

miteshvp:
> I don't agree that this is out of the scope of this PR

What I meant was there are other workers using Graph Functionalities as well. That is out of scope for this PR to change those worker process to start using Graph Adapters.

>My idea was creating class let's say Query. This class would have methods called addVertex and add property. Inside these methods, you could concatenate strings and run a check for injections.

This makes sense if you could generalize things. I have already created string cleaner function as an adapter method, but to generalize to the extent you are saying, does not make sense to me because each task has its own set of properties. Also, it just concatenates the string, and for that, I think we should directly use string concatenation rather than over-doing use of adapter class. WDYT?

humaton:
@miteshvp I have to stand up here. Pavel is right. If we want to achieve consistent well maintainable code, the pattern he describes is the best way to go.

Another thing is that your string operations do not include any tests, that said how do you know that it's working as you meant it?

miteshvp:
> I have to stand up here. Pavel is right. If we want to achieve consistent well maintainable code, the pattern he describes is the best way to go.

Sure. I understand your concerns and Pavel's as well. What I am saying is, we should not over-engineer something. To me it looks like we are almost going in that direction. We should not apply same yardstick to all the problems. Still, feel free to point out if something more can be pulled into adapter codebase. If it makes sense, I will be happy to move it.
cc @pkajaba @fridex  

> Another thing is that your string operations do not include any tests, that said how do you know that it's working as you meant it?

Yet to create one

humaton:
@miteshvp this is no over-engineering this is actually how things are done for future usability. We don't want half-baked solutions.

miteshvp:
@fridex WDYT? I need your suggestions, is there anything else that you think can be part of adapters?

miteshvp:
Test cases are pending. SO marking it as WIP

fridex:
https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/355

tisnik:
needinfo: is this issue still in progress or is it just in a dormant state?

pkajaba:
> needinfo: is this issue still in progress or is it just in a dormant state?

 @miteshvp this question ^^ probably slipped away from you. WDYT?

miteshvp:
Closing this PR. Frido created another one. Any work to shift data-importer onto worker should be handled there. https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/355

PR #338, opened 2017-09-13
[WIP] License filter

harjinder-hari:
This is a **WIP** PR for license filter work.

The task is to avoid recommending those packages that will result in license conflict, if the user chooses that package in her stack. In this work, we need to check if any of companion packages and/or alternate packages may cause a license conflict.

For filtering alternate packages, we have to consider all possible combinations of recommended alternate packages. That is exponential ! ( sum nCi = 2^n with 0 <= i <= n. ) So, we plan to spend more time on figuring out if there is a better way to filter alternate packages.

I have implemented license filter for companion packages at the moment.

Thanks,
Harjinder


f8a_worker/workers/recommender.py
```diff
@@ -607,6 +608,39 @@ def _get_stack_values(self, similar_stacks_list):
             similarity_list.append(s_stack)
         return similarity_list
 
+def apply_license_filter(epv_list):
+    dependencies = []
```
pkajaba: you can do `dependencies = licenses = license_score_list = []`

f8a_worker/workers/recommender.py
```diff
@@ -607,6 +608,39 @@ def _get_stack_values(self, similar_stacks_list):
...
+    licenses = []
+    license_score_list = []
+    for epv in epv_list:
+        name = epv.get('pkg', {}).get('name', [''])[0]
```
pkajaba: you can move the definition of the `name, version and licenses` into the `license_scoring_input` call, can't you?

f8a_worker/workers/recommender.py
```diff
@@ -607,6 +608,39 @@ def _get_stack_values(self, similar_stacks_list):
...
+
+        license_scoring_input = {
+            'package': name,
+            'version': version,
+            'licenses': licenses
+        }
+        license_score_list.append(license_scoring_input)
+
+    # Call License Scoring to Get Stack License
+    license_analysis, dependencies = perform_license_analysis(license_score_list, dependencies)
```
pkajaba: this is really strange... you are using one variable `dependencies` as an argument of the call and then as the return value. as the argument, it's always `[]`...

f8a_worker/workers/recommender.py
```diff
@@ -607,6 +608,39 @@ def _get_stack_values(self, similar_stacks_list):
...
+    conflicting_companions = []
```
pkajaba: This can be set at first place, can't be?

f8a_worker/workers/recommender.py
```diff
@@ -607,6 +608,39 @@ def _get_stack_values(self, similar_stacks_list):
...
+    if license_analysis['status'] == 'StackLicenseConflict':
+        conflict_packages = license_analysis.get('conflict_packages', {})
+        for conflict in conflict_packages:
+            for pkg in conflict.keys():
+                conflicting_companions.append(pkg)
+    conflicting_companions = list(set(conflicting_companions))
```
pkajaba: why do you have to convert this to `list`?

f8a_worker/workers/recommender.py
```diff
@@ -702,8 +736,15 @@ def execute(self, parguments=None):
                                  .format(parguments.get('external_request_id', ''),
                                          set(companion_packages).difference(set(filtered_list))))
 
+                    # Apply License Filters
```
pkajaba: This comment is redundant here since you call method `apply_license_filter` which sort of implies that you are going to filter.

f8a_worker/workers/recommender.py
```diff
@@ -702,8 +736,15 @@ def execute(self, parguments=None):
...
+                    lic_filtered_companions, lic_filtered_list = apply_license_filter(filtered_comp_packages_graph, input_stack)
+
+                    _logger.info("Companion Packages further filtered (based on licenses) for external_request_id {} {}"
```
pkajaba: This line is too long.

f8a_worker/workers/recommender.py
jpopelka: I guess it's to remove duplicates from list

```
In [1]: l=[1,2,1,4,3]

In [2]: list(set(l))
Out[2]: [1, 2, 3, 4]
```

pkajaba: Even though you are right if you check my previous comment  `conflicting_companions` is defined as a `list`. Which is completely unnecessary, since it can be a `set` at first place so we don't have to use this conversion.

f8a_worker/workers/recommender.py
harjinder-hari: DONE. I have clubbed multiple initializations into one line.

f8a_worker/workers/recommender.py
harjinder-hari: DONE. I have moved the definitions.

f8a_worker/workers/recommender.py
harjinder-hari: This was old logic. I have done a new logic that does not require it any more. I plan to close this PR and open a new PR for new logic.

f8a_worker/workers/recommender.py
harjinder-hari: This was part of old logic.

f8a_worker/workers/recommender.py
harjinder-hari: For JSON conversion purposes, I have kept data in list and uses set() for deduplication.

f8a_worker/workers/recommender.py
harjinder-hari: Yes that is right. But I wanted to highlight that next set of lines of code is related to license filters.

f8a_worker/workers/recommender.py
harjinder-hari: Yes, that line is very long. I have now tried best to reduce the length to whatever is currently there.

harjinder-hari:
Thanks @pkajaba for the review ! However, there has been a significant change in the logic and I will open another PR for this work. In that PR, I have addressed your review comments e.g. clubbing multiple initializations into one line, moving definitions of variables, length of lines, etc.

Thanks once again!
-Harjinder

PR #340, opened 2017-09-14
ossindex 'bug'? work-around

jpopelka:
Example:
https://ossindex.net/v2.0/vulnerability/pm/maven/fromtill/1900000000000/-1
should return no results because the timestamp is way in future,
but for some reason there's always at least one entry.

Remove it - if its 'updated' time is lower than the specified 'from_time'.


jpopelka:
[Tests](https://ci.centos.org/job/devtools-fabric8-analytics-worker-fabric8-analytics/86/console) succeeded, but then got stuck somehow.

jpopelka:
[test]

PR #349, opened 2017-09-20
Script to run linter

tisnik:


run-linter.sh
```diff
@@ -0,0 +1,67 @@
+directories="alembic cucoslib tests"
```
jpopelka: `cucoslib` -> `f8a_worker`

PR #351, opened 2017-09-22
collect github-statistics also during manifest data collection

SaketJNU:
Task item for https://github.com/openshiftio/openshift.io/issues/983.
Here, the objective is: when we collect manifest data for most github stared repos., at the sametime we also need to collect watchers count, fork count and star count for the same repos.

### DIFF
--- a/f8a_worker/dispatcher/flows/githubManifestMetadataFlow.yml
+++ b/f8a_worker/dispatcher/flows/githubManifestMetadataFlow.yml
@@ -33,6 +33,6 @@
                    key: 'recursive_limit'
                    value: 0
         - from: 'dependency_snapshot'
-          to:    'github_manifest_details'	
-        - from:    'github_manifest_details'
+          to: 'github_details'
+        - from: 'github_details'
           to: 'GitHubManifestMetadataResultCollector'
--- a/f8a_worker/dispatcher/flows/githubManifestMetadataFlow.yml
+++ b/f8a_worker/dispatcher/flows/githubManifestMetadataFlow.yml
@@ -33,4 +33,6 @@
                    key: 'recursive_limit'
                    value: 0
         - from: 'dependency_snapshot'
+	  to:    'github_manifest_details'	
+	- from:    'github_manifest_details'
           to: 'GitHubManifestMetadataResultCollector'
--- a/f8a_worker/dispatcher/flows/githubManifestMetadataFlow.yml
+++ b/f8a_worker/dispatcher/flows/githubManifestMetadataFlow.yml
@@ -33,6 +33,6 @@
                    key: 'recursive_limit'
                    value: 0
         - from: 'dependency_snapshot'
-	  to:    'github_manifest_details'	
-	- from:    'github_manifest_details'
+          to:    'github_manifest_details'	
+        - from:    'github_manifest_details'
           to: 'GitHubManifestMetadataResultCollector'
--- a/f8a_worker/dispatcher/nodes.yml
+++ b/f8a_worker/dispatcher/nodes.yml
@@ -284,12 +284,6 @@
       max_retry: 0
       storage: 'BayesianPostgres'
       queue: '{DEPLOYMENT_PREFIX}_{WORKER_ADMINISTRATION_REGION}_GitHubManifestMetadataGithubTask_v0'
-    - name: 'github_manifest_details'
-      classname: 'GithubStats'
-      import: 'f8a_worker.workers'
-      max_retry: 0
-      storage: 'BayesianPostgres'
-      queue: '{DEPLOYMENT_PREFIX}_{WORKER_ADMINISTRATION_REGION}_GitHubManifestMetadataGithubTask_v0'
     - name: 'git_stats'
       classname: 'GitStats'
       import: 'f8a_worker.workers'
--- a/f8a_worker/dispatcher/nodes.yml
+++ b/f8a_worker/dispatcher/nodes.yml
@@ -284,6 +284,12 @@
       max_retry: 0
       storage: 'BayesianPostgres'
       queue: '{DEPLOYMENT_PREFIX}_{WORKER_ADMINISTRATION_REGION}_GitHubManifestMetadataGithubTask_v0'
+    - name: 'github_manifest_details'
+      classname: 'GithubStats'
+      import: 'f8a_worker.workers'
+      max_retry: 0
+      storage: 'BayesianPostgres'
+      queue: '{DEPLOYMENT_PREFIX}_{WORKER_ADMINISTRATION_REGION}_GitHubManifestMetadataGithubTask_v0'
     - name: 'git_stats'
       classname: 'GitStats'
       import: 'f8a_worker.workers'
--- a/f8a_worker/workers/githubstats.py
+++ b/f8a_worker/workers/githubstats.py
@@ -1,0 +1,55 @@
+from bs4 import BeautifulSoup
+import requests
+import logging
+from f8a_worker.base import BaseTask
+
+logger = logging.getLogger(__name__)
+
+class GithubStats(BaseTask):
+    """ Collects statistics by parsing Github Page """
+    _analysis_name = "github_manifest_details"
+
+    @classmethod
+    def _get_page(cls, repo_name):
+        github_url = 'https://github.com/'
+        url = github_url + "/" + repo_name
+        raw_data = requests.get(url)
+        if raw_data:
+            soup = BeautifulSoup(raw_data.text, "html.parser")
+            return soup
+        else:
+            logger.info("Could not fetch the page for repo_name:"% repo_name)
+
+    @classmethod
+    def _get_stargazers(cls, soup, repo_name):
+        stars =  soup.find(attrs={"href": repo_name + "/" + "stargazers"})
+        return (stars.text).strip() if stars is not None else 0
+
+    @classmethod
+    def _get_watchers(cls, soup, repo_name):
+        watches =soup.find(attrs={"href": repo_name + "/" + "watchers"})
+        return (watches.text).strip() if watches is not None else 0
+
+    @classmethod
+    def _get_fork_count(cls, soup, repo_name):
+        forks = soup.find(attrs={"href": repo_name + "/" + "network"})
+        return (forks.text).strip() if forks is not None else 0
+
+    @classmethod
+    def execute(cls, arguments):
+        repo_name_passed = arguments.get('repo_name')
+        repo_name = "/" + repo_name_passed
+        ecosystem = arguments.get("ecosystem")
+        soup = cls._get_page(repo_name)
+        if soup is not None:
+            github_stats = {
+                "stars": cls._get_stargazers(soup, repo_name),
+                "forks": cls._get_fork_count(soup, repo_name),
+                "watches": cls._get_watchers(soup, repo_name),
+                "repo_name": repo_name_passed,
+                "ecosystem": ecosystem
+            }
+            return github_stats
+        else:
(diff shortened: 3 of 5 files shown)


f8a_worker/dispatcher/flows/githubManifestMetadataFlow.yml
```diff
@@ -33,4 +33,6 @@
                    key: 'recursive_limit'
                    value: 0
         - from: 'dependency_snapshot'
-          to: 'GitHubManifestMetadataResultCollector'
\ No newline at end of file
+	  to:    'github_manifest_details'	
+	- from:    'github_manifest_details'
```
fridex: Please do not mix tabs and spaces (can be misleading). Use spaces as we already use spaces.

f8a_worker/dispatcher/nodes.yml
```diff
@@ -284,6 +284,12 @@
       max_retry: 0
       storage: 'BayesianPostgres'
       queue: '{DEPLOYMENT_PREFIX}_{WORKER_ADMINISTRATION_REGION}_GitHubManifestMetadataGithubTask_v0'
+    - name: 'github_manifest_details'
+      classname: 'GithubStats'
+      import: 'f8a_worker.workers'
```
fridex: You will need to place the following line to `f8a_worker/workers/__init__.py` to make this import work:
```python
from f8a_worker.workers.githubstats import GithubStats
```

But I would suggest to place this logic to `github_details` task that we already have (see bellow).

f8a_worker/workers/githubstats.py
```diff
@@ -0,0 +1,55 @@
+from bs4 import BeautifulSoup
+import requests
+import logging
+from f8a_worker.base import BaseTask
+
+logger = logging.getLogger(__name__)
```
fridex: Every task you derive from `BaseTask` has already properly instantiated and tweaked logging. You can accept it by accessing `self.log` - e.g. `self.log.info("Info message")`.

f8a_worker/workers/githubstats.py
```diff
@@ -0,0 +1,55 @@
...
+
+class GithubStats(BaseTask):
+    """ Collects statistics by parsing Github Page """
+    _analysis_name = "github_manifest_details"
```
fridex: This is not really needed.

f8a_worker/workers/githubstats.py
```diff
@@ -0,0 +1,55 @@
...
+    @classmethod
+    def _get_page(cls, repo_name):
+        github_url = 'https://github.com/'
+        url = github_url + "/" + repo_name
+        raw_data = requests.get(url)
+        if raw_data:
+            soup = BeautifulSoup(raw_data.text, "html.parser")
+            return soup
+        else:
+            logger.info("Could not fetch the page for repo_name:"% repo_name)
```
fridex: The proper way to handle this would be to raise `FatalTaskError` that gets available after:
```python
from selinon import FatalTaskError
```

If you raise this error system will automatically keep track of failed task (and tasks will not be retried - overrides `retry` configuration option). We have capabilities to retry such failed tasks in future.

f8a_worker/workers/githubstats.py
```diff
@@ -0,0 +1,55 @@
...
+
+    @classmethod
+    def _get_stargazers(cls, soup, repo_name):
+        stars =  soup.find(attrs={"href": repo_name + "/" + "stargazers"})
+        return (stars.text).strip() if stars is not None else 0
```
fridex: Brackets around `stars.text` not needed - here and bellow as well.

f8a_worker/workers/githubstats.py
```diff
@@ -0,0 +1,55 @@
...
+        watches =soup.find(attrs={"href": repo_name + "/" + "watchers"})
+        return (watches.text).strip() if watches is not None else 0
+
+    @classmethod
+    def _get_fork_count(cls, soup, repo_name):
+        forks = soup.find(attrs={"href": repo_name + "/" + "network"})
+        return (forks.text).strip() if forks is not None else 0
+
+    @classmethod
+    def execute(cls, arguments):
```
fridex: Missing strict asserts we put before each task execution that state what is necessary input (sanity checks):
```python
self._strict_assert(arguments.get('repo_name'))
self._strict_assert(arguments.get('ecosystem'))
```

With this and with using instance logging, the `execute` method needs to be directly instance method (not classmethod).

f8a_worker/workers/githubstats.py
```diff
@@ -0,0 +1,55 @@
...
+            github_stats = {
+                "stars": cls._get_stargazers(soup, repo_name),
+                "forks": cls._get_fork_count(soup, repo_name),
+                "watches": cls._get_watchers(soup, repo_name),
+                "repo_name": repo_name_passed,
+                "ecosystem": ecosystem
+            }
+            return github_stats
+        else:
+            logger.info("No Github Statistics is found for repo_name: %s" %repo_name_passed)
```
fridex: Again, raise `FatalTaskResult` here to track this failures in our production logs and keep track of failed tasks so we can (potentially) reschedule them on demand.

f8a_worker/workers/githubstats.py
```diff
@@ -0,0 +1,55 @@
...
+
+    @classmethod
+    def _get_watchers(cls, soup, repo_name):
+        watches =soup.find(attrs={"href": repo_name + "/" + "watchers"})
+        return (watches.text).strip() if watches is not None else 0
+
+    @classmethod
+    def _get_fork_count(cls, soup, repo_name):
+        forks = soup.find(attrs={"href": repo_name + "/" + "network"})
+        return (forks.text).strip() if forks is not None else 0
```
jpopelka: Do these functions return integers or strings ?
If integers, then you need `int(forks.text.strip())`
If strings then you need to return the `0` as `"0"`

f8a_worker/workers/githubstats.py
```diff
@@ -0,0 +1,55 @@
...
+            logger.info("Could not fetch the page for repo_name:"% repo_name)
+
+    @classmethod
+    def _get_stargazers(cls, soup, repo_name):
+        stars =  soup.find(attrs={"href": repo_name + "/" + "stargazers"})
+        return (stars.text).strip() if stars is not None else 0
+
+    @classmethod
+    def _get_watchers(cls, soup, repo_name):
+        watches =soup.find(attrs={"href": repo_name + "/" + "watchers"})
```
jpopelka: Can be simplified a bit: `{"href": repo_name + "/watchers"}`

f8a_worker/workers/githubstats.py
```diff
@@ -0,0 +1,55 @@
...
+logger = logging.getLogger(__name__)
+
+class GithubStats(BaseTask):
+    """ Collects statistics by parsing Github Page """
+    _analysis_name = "github_manifest_details"
+
+    @classmethod
+    def _get_page(cls, repo_name):
+        github_url = 'https://github.com/'
+        url = github_url + "/" + repo_name
```
jpopelka: `github_url` already ends with `/`

Also `github_url` is not used anywhere else so you can squash these two lines into `url = 'https://github.com/' + repo_name`

f8a_worker/workers/githubstats.py
fridex: They should be integers (casting necessary as you've stated).

f8a_worker/dispatcher/nodes.yml
SaketJNU: removed the new task 'github_manifest_details' from node.yml and appended the logic with  'github_details' task.

f8a_worker/workers/githubstats.py
SaketJNU: updated as per comment

f8a_worker/workers/githubstats.py
SaketJNU: Removed the analysis name, as it has also been removed from tasks.

f8a_worker/workers/githubstats.py
SaketJNU: updated the return values as integers.

f8a_worker/workers/githubstats.py
SaketJNU: removed the ending ```/``` form github_url and added with passed repo_name as .

f8a_worker/workers/githubstats.py
SaketJNU: updated w.r.to suggestions.

f8a_worker/workers/githubstats.py
SaketJNU: updated execution function with suggested lines.

f8a_worker/workers/githubstats.py
SaketJNU: added the ```FatalTaskResult``` implementation in execution function.

fridex:
Two things:

If you would like to use this with https://github.com/fabric8-analytics/fabric8-analytics-jobs/pull/171 you will need to store github data on S3 - e.g. by assigning a database adapter to this task (so it will not be stored in postgres like it is done now - storing will probably fail now due to database relations). This database adapter needs to be implemented (see [docs](http://selinon.readthedocs.io/en/latest/storage.html)) if you would like to use it. It should save data somewhere else (new bucket?) so we do not mix package and package-version level data with repo data.

I would suggest to move this gathering logic to github_details so we have all this logic at one place and we can collect these stats also for projects. Also github_stats is too broad, please incorporate keys so they will live along with other relevant keys. A derived class used in this flow would just reuse this logic and place results on S3.

I agree this needs some deeper knowledge, I can help you with these.

fridex:
Ideally we would like to have something called "github as an ecosystem" where we would just pass url to github repo and we would analyse all the dependencies - that would be the best and the most proper design for me, but it is not your task goal...

SaketJNU:
I had a conversation with @msrb  related to this PR and come to know that the implementation for collecting required data-fields (Github-statistics) are already in plcae and working corretly with slightly different filed names, so came to conclusion, that we do't need modification in current logic and job-flow. Thus, this PR can be closed. 
Thanks

fridex:
> I had a conversation with @msrb related to this PR and come to know that the implementation for collecting required data-fields (Github-statistics) are already in plcae and working corretly with slightly different filed names, so came to conclusion, that we do't need modification in current logic and job-flow. Thus, this PR can be closed.
Thanks

What about missing statistics? Will they be placed in github_details task?

SaketJNU:
@fridex, The task already collecting the github statistics and keeping them as a github_details.json in the same folder where dependency_snapshot.json is being stored. The only thing is that field names are different, that we can update in jobs PR during aggregation.

fridex:
> @fridex, The task already collecting the github statistics and keeping them as a github_details.json in the same folder where dependency_snapshot.json is being stored. The only thing is that field names are different, that we can update in jobs PR during aggregation.

OK, closing. Thanks!

miteshvp:
> The task already collecting the github statistics and keeping them as a github_details.json in the same folder where dependency_snapshot.json is being stored. 
@SaketJNU @msrb - please help me understand, whether this github stats are for the main repo or they are for the dependents found inside the main repo? What we are looking at is github_details for the main repo returned by our github query. 
e.g. As part of github search query, if repo returned is <repo_1> which has a pom.xml with dependencies on <epv_1>, <epv_2>, then, 
we would like to capture github statistics for <repo_1> and not for <epv_1> and <epv_2>

msrb:
@miteshvp it's for the main repo, we started capturing GitHub details for the main repo long time ago as we wanted topics for the repo. We collect the details for dependencies as well (if they are hosted on GH and we know that).

miteshvp:
Thanks Michal

PR #354, opened 2017-09-25
Fix 'E722 do not use bare except'

jpopelka:
cc @tisnik


f8a_worker/graphutils.py
```diff
@@ -201,7 +199,7 @@ def get_osio_user_count(ecosystem, name, version):
         response = get_session_retry().post(GREMLIN_SERVER_URL_REST, data=json.dumps(payload))
         json_response = response.json()
         return json_response['result']['data'][0]
-    except:
+    except Exception:
         logger.error("Failed retrieving Gremlin data.")
```
fridex: maybe logger.exception()? (review comment not relevant to this PR change)

PR #363, opened 2017-09-27
Fix several issues with Python security scans

### LINKED ISSUE #131: [Python] false positives in security scans
It has already been [reported](https://github.com/jeremylong/DependencyCheck/issues/694) to DependencyCheck upstream that its experimental analyzer for Python packages has quite a lot of false positives. It is getting CVEs not only for `python-<package>-<version>`, but also for `python-<version>` (Python interpreter itself).

As mentioned in that report I think it might be caused by the wrong `cpe:/a:python:python:<version>` [CPE](https://nvd.nist.gov/products/cpe) identifier used, 

We should either create a [suppression XML](https://jeremylong.github.io/DependencyCheck/general/suppression.html) or check the  DependencyCheck code if there's a possibility how to fix this ourselves - maybe all we need is some ugly but simple hack that would avoid using of these CPEs.

(Copied from https://github.com/baytemp/worker/issues/65)

jpopelka:
Fixes #131


f8a_worker/workers/CVEchecker.py
```diff
@@ -226,24 +230,29 @@ def _python_scan(self, arguments):
...
-                if 'PKG-INFO' in files:
-                    pkg_info = os.path.join(root, 'PKG-INFO')
-                if 'METADATA' in files:
-                    metadata = os.path.join(root, 'METADATA')
-
-            scan_path = egg_info or pkg_info or metadata
+        extracted_tarball = ObjectCache.get_from_dict(arguments).get_extracted_source_tarball()
+        # depcheck needs to be pointed to a specific file, we can't just scan whole directory
+        egg_info = pkg_info = metadata = None
+        for root, dirs, files in os.walk(extracted_tarball):
```
fridex: `dirs` is not used so maybe be explicit with that:
```python
for root, _, files in os.walk(extracted_tarball):
```

tests/workers/test_cvecheck.py
```diff
@@ -119,8 +114,84 @@ def test_python_mako(self):
...
+        #        "severity": "Medium"}]
+        assert isinstance(details, list) and len(details) == 1
+        detail = details[0]
+        assert set(detail.keys()) == {'cvss', 'description', 'id', 'references', 'severity'}
+        cvss = detail['cvss']
+        assert cvss['score'] == 6.8
+        assert cvss['vector'] == 'AV:N/AC:M/Au:?/C:P/I:P/A:P'
+        assert detail['id'] == 'CVE-2015-2296'
+        assert 'http://advisories.mageia.org/MGASA-2015-0120.html' in detail['references']
+        assert detail['severity'] == 'Medium'
```
fridex: Something like [datadiff](https://pypi.python.org/pypi/datadiff) could make this (and the test bellow) more readable and less error prone.

jpopelka:
PTAL

PR #369, opened 2017-10-03
Prefetch 128 messages

fridex:


fridex:
This was for perf tuning purposes. Once this gets to production we should check how does it affect our task flows.

PR #376, opened 2017-10-05
Add support for Golang

msrb:
This is still WIP.

### DIFF
--- a/f8a_worker/solver.py
+++ b/f8a_worker/solver.py
@@ -305,7 +305,7 @@ class GolangReleasesFetcher(ReleasesFetcher):
 
     def fetch_releases(self, package):
         if not package:
-            raise ValueError('package')
+            raise ValueError('package not specified')
 
         package = 'git://{p}.git'.format(p=package)
         output = Git.ls_remote(package, args=['-q'], refs=['HEAD'])
--- a/f8a_worker/solver.py
+++ b/f8a_worker/solver.py
@@ -17,6 +17,7 @@ from urllib.parse import urljoin
 from f8a_worker.enums import EcosystemBackend
 from f8a_worker.models import Analysis, Ecosystem, Package, Version
 from f8a_worker.utils import cwd, tempdir, TimedCommand
+from f8a_worker.process import Git
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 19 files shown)


msrb:
@tuxdna how are we doing with Go support in data-importer?

msrb:
This is in testable state now.

msrb:
Looks like pylint is complaining, @tuxdna could you please take a look?

tuxdna:
@msrb: Looks like these are pylint issues with existing test cases and some other python files. Let me check.

f8a_worker/solver.py
```diff
@@ -298,6 +299,24 @@ def fetch_releases(self, package):
         return package, self.releases_from_maven_org(url)
 
 
+class GolangReleasesFetcher(ReleasesFetcher):
+    def __init__(self, ecosystem):
+        super(GolangReleasesFetcher, self).__init__(ecosystem)
+
+    def fetch_releases(self, package):
+        if not package:
+            raise ValueError('package')
```
fridex: Maybe more descriptive error message?

f8a_worker/utils.py
```diff
@@ -669,5 +670,20 @@ def get_session_retry(retries=3, backoff_factor=0.2, status_forcelist=(404, 500,
     session.mount('http://', adapter)
     return session
 
-# get not hidden files from current directory
-# print(list(get_all_files_from('.', file_filter=lambda a: not startswith(a, ['.']))))
+
+def normalize_package_name(ecosystem, name):
```
fridex: Any particular reason why this is here? I think it should be placed on API server side instead as core should work directly on real package names.

tuxdna:
On invoking: http://localhost:32000/api/v1/component-analyses/go/github.com%2Fstretchr%2Ftestify/master

Its dependencies are getting imported as well now:

```
2017-10-09 13:17:58,794 [INFO] GraphImporterTask: Invoke graph importer at url: 'http://data-model-importer:9192/api/v1/selective_ingest' for {'package_list': [{'version': '0b88a3aa0751a3accc5cadec90e6c9bc88e8219e', 'name': 'github.com/golang/tools', 'ecosystem': 'go'}], 'select_ingest': ['source_licenses', 'digests', 'dependency_snapshot']}
2017-10-09 13:17:59,647 [INFO] GraphImporterTask: Graph import succeeded with response: {
  "count_imported_EPVs": 1,
  "epv": {
    "package_list": [
      {
        "ecosystem": "go",
        "name": "github.com/golang/tools",
        "version": "0b88a3aa0751a3accc5cadec90e6c9bc88e8219e"
      }
    ],
    "select_ingest": [
      "source_licenses",
      "digests",
      "dependency_snapshot"
    ]
  },
  "message": "The import finished successfully!"
}
```

fridex:
[test]

f8a_worker/solver.py
msrb: Good catch.

f8a_worker/utils.py
msrb: Normalizing package names on server side would work just fine for `golang`, but we need to be able to normalize package names in init task for other ecosystem. So I'd keep the logic at one place.

msrb:
PR updated, CI is green.

PR #383, opened 2017-10-06
Make sure alembic can connect to RDS from readiness probe

msrb:


hack/worker-readiness.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+export F8A_POSTGRES=postgresql://${POSTGRESQL_USER}:${POSTGRESQL_PASSWORD}@${PGBOUNCER_SERVICE_HOST:-coreapi-pgbouncer}:${PGBOUNCER_SERVICE_PORT:-5432}/${POSTGRESQL_DATABASE}
```
fridex: What do you think about:
```
F8A_POSTGRES=`python3 -c 'from f8a_worker.defaults import configuration; print(configuration.POSTGRES_CONNECTION)'`
```
This way we place connection string creation at one place and reuse logic.

msrb: Updated 😉

PR #387, opened 2017-10-09
Always resolve POMs in mercator

msrb:
Restore the previous behavior when [[1]](https://github.com/openshiftio/openshift.io/issues/1063) is done/in progress.

### DIFF
--- a/f8a_worker/workers/graphaggregator.py
+++ b/f8a_worker/workers/graphaggregator.py
@@ -61,7 +61,7 @@ class GraphAggregatorTask(BaseTask):
                 # Create instance manually since stack analysis is not handled by dispatcher
                 subtask = MercatorTask.create_test_instance(task_name=self.task_name)
                 arguments['ecosystem'] = manifest['ecosystem']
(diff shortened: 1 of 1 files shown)


f8a_worker/workers/mercator.py
```diff
@@ -193,7 +193,7 @@ def run_mercator(self, arguments, cache_path,
                        'details': []}
         mercator_target = arguments.get('cache_sources_path', cache_path)
         tc = TimedCommand(['mercator', mercator_target])
-        update_env = {'MERCATOR_JAVA_RESOLVE_POMS': 'true'} if resolve_poms else {}
```
jpopelka: Yes, this does the trick, but `resolve_poms` argument becomes unused.

I'd rather see the change in [GraphAggregatorTask](https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/f8a_worker/workers/graphaggregator.py#L64), so that it runs the `MercatorTask` with `resolve_poms=True` (or not specifying `resolve_poms` at all so the default (`True`) value is used)

msrb: Fair enough, update. PTAL.

Issue #386, opened 2017-10-09
ValueError: invalid interpolation syntax in 'postgresql://coreapi:***@bayesian-pgbouncer:5432


msrb:
Failure from prod, seems to be caused by RDS password containing `%` sign. 

```
[...]

PR #388, opened 2017-10-09
Use unquoted postgres password in readiness probe.

pkajaba:
Using quoted version of postgres password caused failure in production.


pkajaba:
Fixes #386

f8a_worker/defaults.py
```diff
@@ -3,6 +3,25 @@
 from urllib.parse import quote
 
 
+def _make_postgres_string(password=''):
+    """
+    Method creates postgres connection string. It's parametrized, so it's possible to create either
+    quoted or unquoted version of connection string.
+    Note that it's outside of class since there is no simple way how to call it inside the class
+    without class initialization.
```
jpopelka: What about marking it as `@staticmethod` and calling it `F8AConfiguration.make_postgres_string()` ?

pkajaba: Chicken or the egg problem. `F8AConfiguration` is not defined in a body of `F8AConfiguration`.

fridex: It would be in the same namespace so something like this should do the trick:

```python
class A(object):

    def _make_foo():
        return "bar"

    FOO = _make_foo()


print(A.FOO)
```

Also note that it needs to be defined before use.

jpopelka: OK, you seem to be right. [Calling a static method from another method works](https://stackoverflow.com/questions/35459360/how-to-call-static-methods-inside-the-same-class-in-python), but we need to call it during initialization of the class, which is a no-go.

pkajaba: Nice, I like that!, I don't even have to expose FOO in my case :-)

PR #400, opened 2017-10-17
fixed wrong number in analyzed_dependencies_count

ravsa:
fixes https://github.com/openshiftio/openshift.io/issues/844

### DIFF
--- a/f8a_worker/workers/stackaggregator_v2.py
+++ b/f8a_worker/workers/stackaggregator_v2.py
@@ -159,14 +159,14 @@ def aggregate_stack_data(stack, manifest_file, ecosystem, deps, manifest_file_pa
             "user_stack_info": {
                 "ecosystem": ecosystem,
                 "analyzed_dependencies_count": len(dependencies),
-                "analyzed_dependencies": deps,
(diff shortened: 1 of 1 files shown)


jpopelka:
[test]

f8a_worker/workers/stackaggregator_v2.py
```diff
@@ -159,14 +159,14 @@ def aggregate_stack_data(stack, manifest_file, ecosystem, deps, manifest_file_pa
             "user_stack_info": {
                 "ecosystem": ecosystem,
                 "analyzed_dependencies_count": len(dependencies),
-                "analyzed_dependencies": deps,
-                "unknown_dependencies_count": 0,
+                "analyzed_dependencies": dependencies,
```
pkajaba: can you rename it correctly? you renamed `deps` to `dependencies` which might lead to same problem in future.

Thanks, it looks good otherwise

jpopelka: I don't see any renaming. Anyway, the commit has been merged as part of #401 so I'm closing this.

PR #407, opened 2017-10-18
Update README.md

jyasveer:
trivial change to trigger cicd


msrb:
I guess this one's not needed any more :)

PR #401, opened 2017-10-17
Fixes issue 'Unknown components' count 

ravsa:
fixes https://github.com/openshiftio/openshift.io/issues/849

PR #410, opened 2017-10-19
Revert "Fixes issue 'Unknown components' count"

fridex:


fridex:
Revert #401. It looks like schema of stack analysis has changed with this PR and e2e tests do not accept this change.

fridex:
I will merge this as unit tests gave it green. If this is not root cause for e2e failures, I will revert this revert.

PR #416, opened 2017-10-23
Utilize and expose S3 object key construction

fridex:


fridex:
Needed for https://github.com/fabric8-analytics/fabric8-analytics-jobs/pull/204

PR #417, opened 2017-10-25
fixes license failure issue

miteshvp:
https://github.com/openshiftio/openshift.io/issues/1213

### DIFF
--- a/f8a_worker/workers/recommender.py
+++ b/f8a_worker/workers/recommender.py
@@ -689,7 +689,7 @@ def apply_license_filter(user_stack_components, epv_list_alt, epv_list_com):
                                                 license_score_list_com)
 
     conflict_packages_alt = conflict_packages_com = []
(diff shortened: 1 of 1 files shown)


f8a_worker/workers/recommender.py
```diff
@@ -689,7 +689,7 @@ def apply_license_filter(user_stack_components, epv_list_alt, epv_list_com):
                                                 license_score_list_com)
 
     conflict_packages_alt = conflict_packages_com = []
-    if la_output['status'] == 'Successful' and la_output['license_filter'] is not None:
+    if la_output.get('status') == 'Successful' and la_output.get('license_filter') is not None:
         license_filter = la_output.get('license_filter', {})
```
jpopelka: This can be changed to `license_filter = la_output['license_filter']` since at this point you are sure that there is a `'license_filter'` key in `la_output`.

PR #421, opened 2017-11-02
Fetch golang packages using go get

tuxdna:
Following component analyses should work fine:

 * http://localhost:32000/api/v1/component-analyses/go/github.com%2fflynn%2fflynn%2fbootstrap/master

Note that the package name is in a sub-folder of the github repo, which is what this PR intends to solve.


tuxdna:
tests related to IndianaJones are failing 

```
tests/test_process.py::TestGit::test_git_add_and_commit_everything_with_dotgit FAILED
tests/test_process.py::TestIndianaJones::test_fetch_npm_specific[abbrev-1.0.7-30f6880e415743312a0021a458dd6d26a7211f803a42f1e4a30ebff44d26b7de] FAILED
tests/test_process.py::TestIndianaJones::test_fetch_npm_specific[abbrev-1.0.4-8dc0f480571a4a19e74f1abd4f31f6a70f94953d1ccafa16ed1a544a19a6f3a8] FAILED
tests/test_process.py::TestIndianaJones::test_fetch_pypi_specific[six-1.0.0-ca79c14c8cb5e58912d185f0e07ca9c687e232b7c68c4b73bf1c83ef5979333e] FAILED
tests/test_process.py::TestIndianaJones::test_fetch_rubygems_specific[permutation-0.1.7-e715cccaccb8e2d1450fbdda85bbe84963a32e9bf612db278cbb3d6781267638] FAILED
tests/test_process.py::TestIndianaJones::test_fetch_maven_specific[com.rabbitmq:amqp-client-3.6.1-cb6cdb7de8d37cb1b15b23867435c7dbbeaa1ca4b766f434138a8b9ef131994f] FAILED
tests/test_process.py::TestIndianaJones::test_fetch_nuget_specific[NUnit-3.7.1-db714c0a01d8a172e6c378144b1192290263f8c308e8e2baba9c11d9fe165db4] FAILED
tests/test_process.py::TestIndianaJones::test_fetch_go_specific[github.com/gorilla/mux-3f19343c7d9ce75569b952758bd236af94956061-50cc6ce3b58fb23f7f4e5ccf8e24897ba63c628fdc4a52ef4648ecdad7a0a0e9] FAILED

```

msrb:
hmm, let's try again

msrb:
[test]

tuxdna:
One test failing `tests/test_solver.py::TestFetcher::test_nuget_fetcher[System.Data.SQLite-expected5] FAILED`

tuxdna:
All tests pass now. Please review.

f8a_worker/process.py
```diff
@@ -440,13 +454,20 @@ def fetch_rubygems_artifact(name, version, target_dir):
...
+        env['GOPATH'] = target_dir
+        TimedCommand.get_command_output(['go', 'get', '-d', name],
+                                        timeout=300,
+                                        env=env,
+                                        graceful=True)
+        package_dir = os.path.join(target_dir, 'src', name)
+        with cwd(package_dir):
+            git = Git(package_dir)
+            git.reset(version, hard=True)
+            artifact_filename = git.archive(version, repo_path='.')
```
jpopelka: I'd guess that `repo_path='.'` does the same as not specifying `repo_path` at all, doesn't it ?

f8a_worker/utils.py
```diff
@@ -271,11 +271,20 @@ def get_command_output(args, graceful=True, is_json=False, **kwargs):
...
             logger.warning("command %s timed out:\n%s", args, ex.output)
+        elif isinstance(ex, CalledProcessError):
+            if ex.returncode == 1 and "no Go files in" in ex.output:  # skip this error
+                logger.warning("No Golang source in base path. Command %s ended with %s\n%s",
+                               args, ex.returncode, ex.output)
+            else:
+                logger.warning("command %s ended with %s\n%s", args, ex.returncode, ex.output)
         else:
-            logger.warning("command %s ended with %s\n%s", args, ex.returncode, ex.output)
+                logger.warning("command %s ended with %s\n%s", args, ex.returncode, ex.output)
```
jpopelka: Looks like too many indentation spaces (it should start on the same column as the colon on line above it).

Anyway, we catch only `CalledProcessError` and `TimeoutExpired`, so this `else:` is dead code (should never happen).

f8a_worker/utils.py
```diff
@@ -271,11 +271,20 @@ def get_command_output(args, graceful=True, is_json=False, **kwargs):
...
+
         if not graceful:
             logger.error("exception is fatal")
             raise TaskError("Error during running command %s: %r" % (args, ex.output))
+        else:
+            logger.error("Ignoring because graceful flag is set")
```
jpopelka: Then it's not an `error`, but just a `debug`, right ?

f8a_worker/utils.py
```diff
@@ -271,11 +271,20 @@ def get_command_output(args, graceful=True, is_json=False, **kwargs):
         #  (while not mixing it with stdout that is returned if the subprocess succeeds)
         if isinstance(ex, TimeoutExpired):
             logger.warning("command %s timed out:\n%s", args, ex.output)
+        elif isinstance(ex, CalledProcessError):
+            if ex.returncode == 1 and "no Go files in" in ex.output:  # skip this error
+                logger.warning("No Golang source in base path. Command %s ended with %s\n%s",
+                               args, ex.returncode, ex.output)
```
jpopelka: Frankly, I don't like this `Go` specific code in such a generic function as `get_command_output()`.
I guess you've added it due to the calling `go get`, but you call it with `graceful=True`, so any errors are silenced anyway.

jpopelka:
Also please remove the latest commit, since Frido has already merged my #422

f8a_worker/utils.py
tuxdna: Correct, fixing it.

f8a_worker/utils.py
tuxdna: Fixing it.

f8a_worker/process.py
tuxdna: @jpopelka The default value of `repo_path` parameter is `None`. Not sure if `with cwd(self.repo_path)` will fail with that value. @msrb: Could you comment on this ?

f8a_worker/utils.py
tuxdna: @jpopelka You are right, this is not a good design, but it works for now. Will try to refactor this part.

f8a_worker/process.py
jpopelka: It won't fail because there's `repo_path` argument and `self.repo_path` attribute.
Those two are different things, hence I'd suggest renaming `repo_path` argument.

f8a_worker/process.py
```diff
@@ -148,12 +148,26 @@ def add_and_commit_everything(self, message="blank"):
...
         self.commit(message=message)
 
-    def archive(self, basename):
+    def archive(self, basename, repo_path=None):
+        """
+        Create an archive; simply calls `git archive`.
+
+        :param basename: str, name of the resulting archive, without file extension (suffix)
+        :param repo_path: str, only add files found under this path to the archive;
+                          default: add all files from the repository (.git/ is always excluded)
```
jpopelka: I'd suggest using another name (like just `path`), because there already is `self.repo_path`.
`path` is subdirectory (or file) in `self.repo_path`.

f8a_worker/process.py
msrb: Ah, good catch. @jpopelka is right. `repo_path` from `archive()` method overshadows `self.repo_path` object attribute. It works, but it's confusing. We should rename the method param.

msrb:
@tuxdna I haven't tested the code yet, but I think `dependency_snapshot` worker will be still failing (?)

tuxdna:
[test]

tuxdna:
@jpopelka All tests are passing now, and I have fixed code as per your review comments. Please review.


@msrb In my test runs, on invoking a scan for `github.com/flynn/flynn/bootstrap` using the url:

  * http://localhost:32000/api/v1/component-analyses/go/github.com%2fflynn%2fflynn%2fbootstrap/e0d6f32bfccc50bf65c8a628caed2a2aadb79255

I see its dependencies are also being scanned, and so forth. Do you think `dependency_snapshot` should also work given that there are no failures reported in logs ?

jpopelka:
I have no other objections.
@msrb PTAL & merge

tuxdna:
@msrb, @jpopelka : Thanks for your review and approving this merge request!

PR #423, opened 2017-11-02
Bump base image to get latest mercator

msrb:


msrb:
Pulls in https://github.com/fabric8-analytics/mercator-go/pull/20.

PR #424, opened 2017-11-02
Collect license information from GitHub in githuber worker

pkajaba:


pkajaba:
@msrb do we need license name or  spdx_id?

GitHub returns this:
`{'key': 'bsd-3-clause', 'name': 'BSD 3-clause "New" or "Revised" License', 'spdx_id': 'BSD-3-Clause', 'url': 'https://api.github.com/licenses/bsd-3-clause', 'featured': False}`

f8a_worker/workers/githuber.py
```diff
@@ -87,6 +87,23 @@ def _get_topics(self):
...
+        :return: name of repository license
+        """
+        if not self._repo_name:
+            return []
+
+        # GitHub returns always just one license per repository
+        license = requests.get(
+            "https://api.github.com/repos/{repo_name}/license".format(repo_name=self._repo_name)).json()
+        if license['message'] == 'Not Found':
+            return []
```
fridex: Why are you returning array here and above? If I understand it correctly, github returns a string (one license string)?

pkajaba: great catch :-D I copy-pasted topics and did not think that much about it. I will fix it

f8a_worker/workers/githuber.py
```diff
@@ -87,6 +87,24 @@ def _get_topics(self):
...
+        Get license from GitHub repository through GitHub REST API
+
+        :return: name of repository license
+        """
+        if not self._repo_name:
+            return ""
+
+        # GitHub returns always just one license per repository
+        license = requests.get(
+            "https://api.github.com/repos/{repo_name}/license".format(
```
msrb: Do we really need the extra API call? I don't know, but this means we will hit the API limit much sooner.

pkajaba: Does it count even if I don't use a token for this query? I will try to look whether there not license information inside query which we are doing in this worker.

msrb: Good point, I think all `api.github.com` calls without authentication are limited to 60 per hour. So this would be definitely no go for us.

f8a_worker/workers/schemas/github_details.py
```diff
@@ -8,12 +8,14 @@
...
+ROLE_v1_0_5 = "v1-0-5"
 ROLE_v2_0_0 = "v2-0-0"
 ROLE_TITLE = jsl.roles.Var({
     ROLE_v1_0_0: "Github Results v1-0-0",
     ROLE_v1_0_1: "Github Results v1-0-1",
     ROLE_v1_0_3: "Github Results v1-0-3",
     ROLE_v1_0_4: "Github Results v1-0-4",
+    ROLE_v1_0_5: "Github Results v1-0-5",
     ROLE_v2_0_0: "Github Results v2-0-0",
 })
```
jpopelka: Why `1-0-5` and not `2-0-1` ? Looks like @msrb added the `2-0-0` with 5c46e8b, but didn't bump the used schema version in `GithubTask`. @msrb do you remember whether it was on purpose or by mistake ?

msrb:
> @msrb do we need license name or spdx_id?

I'd take the whole `license` dict returned by the API.

f8a_worker/workers/schemas/github_details.py
msrb: It was a mistake.

pkajaba:
@msrb about GH API discussion: 

You are right that they state in their documentation https://developer.github.com/v3/licenses/#get-a-repositorys-license, but try to use this command in your terminal:
`curl https://api.github.com/repos/almighty/almighty-jobs`
or 
`curl -H "Accept application/vnd.github.drax-preview+json" https://api.github.com/repos/almighty/almighty-jobs`
(header is specified in documentation https://developer.github.com/v3/licenses/#get-a-repositorys-license)

On the other hand, this returns license:
`curl https://api.github.com/repos/almighty/almighty-jobs/license`

So my conclusion here is that their API documentation is not up-to-date.

pkajaba:
I was missing `:` after `Accept`. Thanks for pointing out @msrb

pkajaba:
@msrb I have decided to replace PyGithub with plain requests calls.

My reasoning: We can get repo data, licenses, and topics with one API call. I could send PR for PyGithub library for a license (there is one for topics [1]), but this would cost 3 calls instead of 1.

[1] https://github.com/PyGithub/PyGithub/pull/635

jpopelka:
> I have decided to replace PyGithub with plain requests calls.

One useful thing that PyGithub does is that it [follows redirected repos](https://github.com/PyGithub/PyGithub/blob/master/github/Requester.py#L289), see

```python
import github
github.Github().get_repo('docker/docker').raw_data.get('url')
'https://api.github.com/repos/moby/moby'
```

So if you want to do that change I'd like to see you follow redirects too
```
curl https://api.github.com/repos/docker/docker
{
  "message": "Moved Permanently",
  "url": "https://api.github.com/repositories/7691631",
  "documentation_url": "https://developer.github.com/v3/#http-redirects"
}
curl https://api.github.com/repositories/7691631
{
  ...
  "url": "https://api.github.com/repos/moby/moby"
  ...
}
```

pkajaba:
> One useful thing that PyGithub does is that it follows redirected repos

That's a great catch! Thanks for pointing out. I will implement this functionality.

pkajaba:
[test]

f8a_worker/workers/githuber.py
```diff
@@ -28,41 +37,41 @@ def create_test_instance(cls, repo_name, repo_url):
...
+        except requests.exceptions.HTTPError as err:
+            self.log.error(
+                "Failed to get results from {url} with {err}".format(url=url, err=err)
+            )
+            return {}
 
     @classmethod
     def _get_last_years_commits(cls, repo):
-        activity = cls._retry_no_cached(repo.get_stats_commit_activity)
+        activity = self._get_response(urljoin(repo_url + '/', "stats/commit_activity"))
```
msrb: This will fail as this is class method and there is no `repo_url` defined in this scope.

f8a_worker/workers/githuber.py
```diff
@@ -28,41 +37,40 @@ def create_test_instance(cls, repo_name, repo_url):
...
+                if response:
+                    return response
+                time.sleep(sleep_time)
+        except requests.exceptions.HTTPError as err:
+            self.log.error(
+                "Failed to get results from {url} with {err}".format(url=url, err=err)
+            )
+            return {}
+
+    def _get_last_years_commits(cls, repo_url):
```
msrb: s/cls/self/

msrb:
Ah, this is what's wrong with the tests: https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/tests/workers/test_githuber.py#L22-L23

f8a_worker/workers/githuber.py
```diff
@@ -28,41 +37,40 @@ def create_test_instance(cls, repo_name, repo_url):
...
+                response.raise_for_status()
+                response = response.json()
+                if response:
+                    return response
+                time.sleep(sleep_time)
+        except requests.exceptions.HTTPError as err:
+            self.log.error(
+                "Failed to get results from {url} with {err}".format(url=url, err=err)
+            )
+            return {}
```
fridex: Maybe raise here an exception so caller can easily distinguish an error state.

f8a_worker/workers/githuber.py
```diff
@@ -28,41 +37,40 @@ def create_test_instance(cls, repo_name, repo_url):
...
+
```
fridex: I would expect that there would be raised an exception if retry count is reached and no results are available.

f8a_worker/workers/githuber.py
```diff
@@ -100,8 +95,8 @@ def execute(self, arguments):
                 return result_data
 
         token = self.configuration.GITHUB_TOKEN
-        if not token:
-            if self._rate_limit_exceeded(github.Github()):
+        if token == 'not-set':
```
fridex: I would rather like to have a method in the configuration class for this not to spread these specific values in sources.

f8a_worker/workers/githuber.py
```diff
@@ -114,12 +109,10 @@ def execute(self, arguments):
         else:
             # there might be more comma-separated tokens, randomly select one
             token = random.choice(token.split(',')).strip()
```
fridex: Maybe a method for this in the configuration as well so it is done at one place consistently.

tests/workers/test_githuber.py
```diff
@@ -19,16 +19,21 @@ def test_execute(self, repo_name, repo_url):
         assert results is not None
         assert isinstance(results, dict)
         assert set(results.keys()) == {'details', 'status', 'summary'}
-        if not results['details'].keys():  # no github api token provided
-            return
```
fridex: What if the token is not provided? Maybe [mark the test as skipped](https://docs.pytest.org/en/latest/skipping.html)?

msrb:
[test]

tests/workers/test_githuber.py
pkajaba: token is not provided by default in tests. It might fail because of rate limit, but we decided to give a shot to this approach since we should not exceed 60 requests per hour in tests.

f8a_worker/workers/githuber.py
```diff
@@ -99,27 +74,19 @@ def execute(self, arguments):
...
+        try:
+            _, header = self.configuration.select_random_github_token()
+            self._headers.update(header)
+        except F8AConfigurationException as e:
+            self.log.error(e)
+            result_data['status'] = 'error'
+            return result_data
 
-        gh = github.Github(login_or_token=token)
+        repo_url = (urljoin(self.configuration.GITHUB_API + "repos/", self._repo_name))
```
fridex: Brackets not needed.

f8a_worker/defaults.py
```diff
@@ -111,5 +119,40 @@ def is_local_deployment(cls):
...
+
+    @classmethod
+    def _decide_token_usage(cls):
+        if cls.GITHUB_TOKEN == 'not-set':
+            logger.warning("No Github API token provided (GITHUB_TOKEN env variable), "
+                           "requests will be unauthenticated i.e. limited to 60 per hour")
+            return None
+        else:
+            # there might be more comma-separated tokens, randomly select one
+            return random.choice(cls.GITHUB_TOKEN.split(',')).strip()
```
fridex: Split could be done once as optimization, but that is nitpicking. It's OK :)

f8a_worker/utils.py
```diff
@@ -690,3 +691,28 @@ def get_user_email(user_profile):
...
+            response = response.json()
+            if response:
+                return response
+            time.sleep(sleep_time)
+        else:
+            raise TaskError("Number of retries exceeded")
+    except requests.exceptions.HTTPError as err:
+        message = "Failed to get results from {url} with {err}".format(url=url, err=err)
+        logger.error(message)
+        raise TaskError(message)
```
fridex:
```python
raise TaskError(message) from err
```

f8a_worker/utils.py
```diff
@@ -690,3 +691,28 @@ def get_user_email(user_profile):
...
+
+def get_response(url, headers=None, sleep_time=2, retry_count=10):
+    """
+    Simple wrapper for requests which tries to get response
+    :param url: URL where to do the request
+    :param headers: additional headers for request
+    :param sleep_time: sleep time between retries
+    :param retry_count: number of retries
+    :return: content of response's json
+    """
```
fridex: According to pep for docstring, this should be:
```python
   """Simple wrapper for requests which tries to get response.

   :param url: URL where to do the request
   :param headers: additional headers for request
   :param sleep_time: sleep time between retries
   :param retry_count: number of retries
   :return: content of response's json
   """
```

(If we want to start using proper docstrings that pass checks of tools such as pydocstyle).

fridex:
[test]

PR #426, opened 2017-11-03
Create index on WorkerResult.external_request_id

msrb:


tisnik:
It looks ok. Thanks a lot!

PR #439, opened 2017-11-13
Update nodes.yml

jyasveer:
Updated the bucket name to prod-* for two reasons :-
1. To lessen the burden of bucket creation in s3
2. As this bucket will be a read only bucket.

### DIFF
--- a/f8a_worker/dispatcher/nodes.yml
+++ b/f8a_worker/dispatcher/nodes.yml
@@ -477,7 +477,7 @@
       import: 'f8a_worker.storages'
       configuration:
         <<: *configurationS3
-        bucket_name: '{DEPLOYMENT_PREFIX}-manual-tags'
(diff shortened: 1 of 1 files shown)


f8a_worker/dispatcher/nodes.yml
```diff
@@ -477,7 +477,7 @@
       import: 'f8a_worker.storages'
       configuration:
         <<: *configurationS3
-        bucket_name: '{DEPLOYMENT_PREFIX}-manual-tags'
+        bucket_name: 'prod-manual-tags'
```
fridex: Are you sure with this? This can introduce a lot of harm as devs from dev or staging cluster can write to production buckets.

jyasveer: yes @fridex i had a discussion with @miteshvp and @samuzzal-choudhury regarding this.

PR #441, opened 2017-11-14
Update nodes.yml

jyasveer:
updating it to the prod bucket instance to read master tags

### DIFF
--- a/f8a_worker/dispatcher/nodes.yml
+++ b/f8a_worker/dispatcher/nodes.yml
@@ -477,7 +477,7 @@
       import: 'f8a_worker.storages'
       configuration:
         <<: *configurationS3
-        bucket_name: '{DEPLOYMENT_PREFIX}-manual-tags'
+        bucket_name: 'prod-manual-tags'
         region_name: 'us-east-1'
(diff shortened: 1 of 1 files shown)


msrb:
Question: who and when populates the bucket?

jyasveer:
bucket gets populated by analytics team when they run the offline process to validate the tags gathered from the crowd sourcing.
Server api will use this bucket only for read purpose.

msrb:
[test]

msrb:
@miteshvp @samuzzal-choudhury I will leave this one up to you guys, as you understand the problem better. But we should be careful here, make sure that random deployment on the dev cluster won't override production data.

miteshvp:
@msrb - I agree your concerns are all valid. But we may have to trade-off between duplicating vs. shared resources. Just so that I could be of help, this bucket is not used by any analytics but rather published by them so that API can use them. SO over-all I guess we should be good, even if API manipulates it, we could get it back easily.

jyasveer:
@miteshvp @samuzzal-choudhury can onyone of you please merge this PR ?

PR #443, opened 2017-11-16
Add codecov for unit tests

fridex:


fridex:
CC: @tisnik

Dockerfile.tests
```diff
@@ -7,7 +7,8 @@ WORKDIR /f8a_worker
 RUN cd tests && pip3 install -r ./requirements.txt
 
 RUN yum install -y --setopt=tsflags=nodocs mock && yum clean all && \
+    pip3 install pytest-cov && \
```
jpopelka: This should go into `tests/requirements.in/txt` I think.

Dockerfile.tests
```diff
@@ -7,7 +7,8 @@ WORKDIR /f8a_worker
...
     useradd -o -u 1000 -G mock f8a
-USER f8a
+# USER f8a
 
-CMD ["py.test"]
+CMD ["py.test", "--cov=/f8a_worker/f8a_worker"]
```
jpopelka: `--cov=f8a_worker` should be enough since `WORKDIR /f8a_worker`

Dockerfile.tests
```diff
@@ -7,7 +7,8 @@ WORKDIR /f8a_worker
 RUN cd tests && pip3 install -r ./requirements.txt
 
 RUN yum install -y --setopt=tsflags=nodocs mock && yum clean all && \
+    pip3 install pytest-cov && \
     useradd -o -u 1000 -G mock f8a
-USER f8a
+# USER f8a
```
jpopelka: Why do you think we need this ? It seems to work as it is (i.e. uncommented)

runtest.sh
```diff
@@ -86,7 +86,7 @@ docker run \
 
 echo "Starting test suite"
 docker run -t \
-  -v "${here}:/f8a_worker:ro,Z" \
+  -v "${here}:/f8a_worker:rw,Z" \
```
jpopelka: This is unfortunate, but I don't have any better solution, so I think I can live with that if it's just for tests.
Maybe if you could put a comment there, explaining it ?

fridex:
Thanks @jpopelka for such review! This was just WIP hack-ish after-lunch patch for Pavel to check codecov results (thus marked as WIP) . I can rework it to make it merge-able though.

To reply to review comments: codecov plugin for pytest creates a temporary file `.codecov` inside directory - that's why I mounted volume as rw and commented out user to run tests under root. I couldn't quickly find a way how to print results just to console (not stated in the docs AFAIK). I will rework it to place it somewhere more suitable and writeable (e.g. `/tmp`).

fridex:
OK, so it is not possible to explicitly specify path in codecov (see https://bitbucket.org/ned/coveragepy/src/1a67de1b8c5831365bdf563fe22a607b2b75a4f4/coverage/config.py?at=default&fileviewer=file-view-default#config.py-451 - it expects the file to be actually present if stated explicitly).

I changed workdir to be /tmp so it is writeable. @jpopelka PTAL if @tisnik wants coverage support.

jpopelka:
@tisnik do we want to merge this ?

tisnik:
Perfect! TYVM!

PR #449, opened 2017-11-20
Fix bug - ValueError: need more than 2 values to unpack

tuxdna:
Fixes issue: https://github.com/openshiftio/openshift.io/issues/1394

### DIFF
--- a/f8a_worker/solver.py
+++ b/f8a_worker/solver.py
@@ -307,9 +307,18 @@ class GolangReleasesFetcher(ReleasesFetcher):
         if not package:
             raise ValueError('package not specified')
 
-        gh_host, gh_org, gh_proj = package.split("/")[:3]
-        repo_url = 'git://{gh_host}/{gh_org}/{gh_proj}.git'.format(
-            gh_host=gh_host, gh_org=gh_org, gh_proj=gh_proj)
+        print("Input package: %s" % package)
+
+        parts = package.split("/")[:3]
+        if len(parts) == 3:  # this assumes github.com/org/project like structure
+            host, org, proj = parts
+            repo_url = 'git://{host}/{org}/{proj}.git'.format(host=host, org=org, proj=proj)
+        elif len(parts) == 2 and parts[0] == 'gopkg.in':  # specific to gopkg.in/packages
+            host, proj = parts
+            repo_url = 'https://{host}/{proj}.git'.format(host=host, proj=proj)
+        else:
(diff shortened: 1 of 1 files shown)


tuxdna:
Please review

f8a_worker/solver.py
```diff
@@ -307,9 +307,18 @@ def fetch_releases(self, package):
         if not package:
             raise ValueError('package not specified')
 
-        gh_host, gh_org, gh_proj = package.split("/")[:3]
-        repo_url = 'git://{gh_host}/{gh_org}/{gh_proj}.git'.format(
-            gh_host=gh_host, gh_org=gh_org, gh_proj=gh_proj)
+        print("Input package: %s" % package)
```
jpopelka: Looks like left-over debug message.
You can use  `logger.debug()` if you need it, but make the message more descriptive.

f8a_worker/solver.py
```diff
@@ -307,9 +307,18 @@ def fetch_releases(self, package):
...
+
+        parts = package.split("/")[:3]
+        if len(parts) == 3:  # this assumes github.com/org/project like structure
+            host, org, proj = parts
+            repo_url = 'git://{host}/{org}/{proj}.git'.format(host=host, org=org, proj=proj)
+        elif len(parts) == 2 and parts[0] == 'gopkg.in':  # specific to gopkg.in/packages
+            host, proj = parts
+            repo_url = 'https://{host}/{proj}.git'.format(host=host, proj=proj)
+        else:
+            raise ValueError("Package {} is not supported for ingestion".format(package))
```
jpopelka: That's quite high-level message, what about something like: *"Package {} is invalid git repository"*

msrb:
@tuxdna what would it take to actually start supporting all golang packages backed by git SCM? I know golang packages can be use mercurial and other SCMs, but my guess is that vast majority actually uses git.

tuxdna:
@msrb: For supporting Go packages backed by all ecosystems there are ( AFIAK ) two important steps:

 * Step 1: `go get -d` 
 * Step 2: `git ls-remote`

Step 1 takes care of downloading the package irrespective of the SCM.

It is the Step 2 that is specific to git. Why do we do that? To find latest version of the package ( dependency snapshot ). We have to work-around the Step 2 to prepare dependency snapshots, and then we can be independent of any SCM. Let me create a separate issue for this.

tuxdna:
@jpopelka Please review.

tuxdna:
@jpopelka Thanks for the review and merging this PR.

PR #450, opened 2017-11-20
Bump base image

msrb:


msrb:
To get the new mercator RPM.

msrb:
Merging with thumbs up from @fridex and Jenkins. Thanks 😉

PR #461, opened 2017-11-30
Fix "TypeError: utcnow() takes no arguments (1 given)"

msrb:


fridex:
The CI failure is not relevant. It will need changes in the github tests.

PR #464, opened 2017-11-30
GitHub License API is stable now

msrb:


msrb:
https://developer.github.com/changes/2017-11-28-ending-the-license-api-preview/

PR #465, opened 2017-12-06
use declared_licenses instead of licenses (detected)

miteshvp:
Currently we show detected `licenses` on OSIO dashboard and license scoring also uses detected licenses to calculate stack license. It should be modified to use `declared_licenses`.

### DIFF
--- a/f8a_worker/graphutils.py
+++ b/f8a_worker/graphutils.py
@@ -137,7 +137,7 @@ def extract_component_details(component):
             "vulnerabilities": cves
         }
 
-    licenses = component.get("version", {}).get("licenses", [])
+    licenses = component.get("version", {}).get("declared_licenses", [])
     name = component.get("version", {}).get("pname", [""])[0]
     version = component.get("version", {}).get("version", [""])[0]
     ecosystem = component.get("version", {}).get("pecosystem", [""])[0]
@@ -213,7 +213,7 @@ def create_package_dict(graph_results, alt_dict=None):
                 'ecosystem': ecosystem,
                 'name': name,
                 'version': version,
-                'licenses': epv['ver'].get('licenses', []),
+                'licenses': epv['ver'].get('declared_licenses', []),
                 'latest_version': select_latest_version(
                     epv['pkg'].get('libio_latest_version', [''])[0],
                     epv['pkg'].get('latest_version', [''])[0]),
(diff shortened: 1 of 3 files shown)


msrb:
Fix for the test failures is here: https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/466

msrb:
Detected licenses are, in my opinion, much more valuable than declared licenses. If people really care about licensing, they need to care about what's really in the code, not what upstream declares.

What's the reason behind this change?

miteshvp:
@msrb - I guess it was already discussed to use declared licenses. cc @krishnapaparaju Please let us know if there is any change.

msrb:
@miteshvp yeah, I remember that we discussed how to go about ecosystems for which we won't have detected licenses, because artifacts are only available in binary form (.NET for example).

On the other hand, libraries in Golang usually don't have any manifest files so most of them won't have any declared license. We take licenses reported by GitHub and we have detected licenses.

miteshvp:
@msrb looks like a cabal agenda. Adding it

miteshvp:
[test]

msrb:
> @msrb looks like a cabal agenda. Adding it

Good idea, thanks 😉

msrb:
@miteshvp what's the plan here? you want to merge this now, or you want to wait for the cabal?

miteshvp:
let's wait for the cabal. Thanks.

pombredanne:
FWIW, ScanCode progressively also detects (and will provide a normalized license expression) for the "declared license". I call these "asserted" but that the same. https://github.com/nexB/scancode-toolkit/blob/275-streamline-package-manifests-models/src/packagedcode/models.py#L509

Also on the topic of Windows DLLs, there are two things there:
1. when in a nuget, there is a license URL in the nuspec or API data. These are progressively collected in ScanCode as detection rules to map these to a proper declared or asserted license.
2. DLLs have metadata such as a copyrights and sometimes licenses. If you scan these (which is somewhat heavy) these metadata will be detected soon when this branch is merged: https://github.com/nexB/scancode-toolkit/blob/275-streamline-package-manifests-models/src/packagedcode/win_pe.py

@msrb you also wrote:

> Detected licenses are, in my opinion, much more valuable than declared licenses. If people really care about licensing, they need to care about what's really in the code, not what upstream declares.

I could not agree more, but both are worth IMHO. But in some areas such as NPMs that I think defaults to have an ISC license when none is set, there is a huge amount of discrepancies in these cases between the "declared" license as found in a package.json or regsitry API call and the actual license detected in the code.
My 2 cents is to always take a package.json ISC as not something to put in the bank.

miteshvp:
Thanks @pombredanne for a detailed comment. It really helps and will consider NPM case accordingly.

jpopelka:
Also related:  https://github.com/fabric8-analytics/fabric8-analytics-license-check/issues/9

miteshvp:
So, @msrb - we discussed this in cabal and we will go with `declared_license` and gradually will move to `detected_license` for some ecosystem and `declared_license` for some or in combination of both later.

PR #471, opened 2017-12-09
Stream files from disk while uploading them to S3

msrb:
Service with 2Gi of memory is getting OOM killed while uploading 650Mi
file.

### DIFF
--- a/f8a_worker/storages/s3.py
+++ b/f8a_worker/storages/s3.py
@@ -154,13 +154,19 @@ class AmazonS3(DataStorage):
         :param object_key: object key under which the file should be stored
         :return: object version or None if versioning is off
         """
-        with open(file_path, 'rb') as f:
-            return self.store_blob(f.read(), object_key)
+
+        f = None
+        try:
+            f = open(file_path, 'rb')
+            return self.store_blob(f, object_key)
+        finally:
+            if f is not None:
+                f.close()
 
     def store_blob(self, blob, object_key):
         """ Store blob on S3
 
-        :param blob: bytes to be stored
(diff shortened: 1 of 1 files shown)


msrb:
This is needed to make https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/471 finally work.

msrb:
Tested, merging.

f8a_worker/storages/s3.py
```diff
@@ -154,13 +154,19 @@ def store_file(self, file_path, object_key):
...
+            return self.store_blob(f, object_key)
+        finally:
+            if f is not None:
+                f.close()
 
     def store_blob(self, blob, object_key):
         """ Store blob on S3
 
-        :param blob: bytes to be stored
+        :param blob: bytes or stream to be stored
```
jpopelka: [S3.Object.put() doc](http://boto3.readthedocs.io/en/latest/reference/services/s3.html#S3.Object.put) says it's `bytes or seekable file-like object`, which I like more, but I guess this is OK too.

f8a_worker/storages/s3.py
```diff
@@ -154,13 +154,19 @@ def store_file(self, file_path, object_key):
...
-        with open(file_path, 'rb') as f:
-            return self.store_blob(f.read(), object_key)
+
+        f = None
+        try:
+            f = open(file_path, 'rb')
+            return self.store_blob(f, object_key)
+        finally:
+            if f is not None:
+                f.close()
```
jpopelka: Looks OK, but why do you think the `with` statement would not work any more ?

PR #479, opened 2017-12-18
[WIP]: Add key-value for component companion count

sara-02:
Part of user-story https://github.com/openshiftio/openshift.io/issues/assigned/sara-02

### DIFF
--- a/f8a_worker/graphutils.py
+++ b/f8a_worker/graphutils.py
@@ -223,6 +223,12 @@ def create_package_dict(graph_results, alt_dict=None):
             }
             if epv['pkg'].get('cooccurrence_probability'):
                 pkg_dict['cooccurrence_probability'] = epv['pkg']['cooccurrence_probability']
+
+            # Add the co-occurence count as well.
+            if epv['pkg'].get('cooccurrence_count'):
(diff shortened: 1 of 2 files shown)


f8a_worker/graphutils.py
```diff
@@ -223,6 +223,12 @@ def create_package_dict(graph_results, alt_dict=None):
             }
             if epv['pkg'].get('cooccurrence_probability'):
                 pkg_dict['cooccurrence_probability'] = epv['pkg']['cooccurrence_probability']
+
+            # Add the co-occurence count as well.
+            if epv['pkg'].get('cooccurrence_count'):
```
pkajaba: `pkg_dict['cooccurrence_count'] = epv['pkg'].get('cooccurrence_count', 0)` 

This is much better in context of usage in `recommender.py` Can you change it? @sara-02

pkajaba: sure :-) I completely understand, but still, would you like to open new PR?

sara-02: @pkajaba Sure, will add a new PR and  fix the issue in other places in this file.

PR #486, opened 2017-12-21
Bump worker-base image to pull in new mercator with gradle support

jpopelka:


jpopelka:
passed, merging myself

PR #492, opened 2018-01-09
Add feedback schema model

sara-02:
For issue https://github.com/openshiftio/openshift.io/issues/1740


sara-02:
Do not merge, need to add a ecosystem field as well. Querying the feedback per ecosystem will become easier.

sara-02:
@msrb I have added the relations as per my understanding, please take a look.

f8a_worker/models.py
```diff
@@ -388,3 +388,14 @@ class DownstreamMap(Base):
...
+
+class RecommendationFeedback(Base):
+    __tablename__ = "recommendation_feedback"
+
+    id = Column(Integer, autoincrement=True, primary_key=True)
+    stack_id = Column(String(64), nullable=False)
+    package_name = Column(String(255), nullable=False)
+    recommendation_type = Column(String(255), nullable=False)
+    feedback_type = Column(Boolean, nullable=False, default=False)
+    ecosystem = Column(String(64), nullable=False)
```
msrb: Shouldn't `ecosystem` be a foreign key pointing to `StackAnalysisRequest` table (?) And the same for `package_name`?

sara-02: @msrb this is where I need help. Ecosystem will be a foreign key and the Stackid will also be a ForeignKey. 
Package Name should not be because we will always have a delta between the data synced at the graph and the one for PGM training. So we can let package name, but we need relations for ecosystem and Stackid. I could not figure how to add those.

sara-02:
@msrb I have added the relations, PTAL :)

f8a_worker/models.py
```diff
@@ -111,13 +115,15 @@ def by_name(cls, session, name):
 class Package(Base):
     __tablename__ = 'packages'
     # ecosystem_id together with name must be unique
-    __table_args__ = (UniqueConstraint('ecosystem_id', 'name', name='ep_unique'),)
+    __table_args__ = (UniqueConstraint(
```
msrb: Please don't mix new code and formatting changes in a single commit. It makes it harder to review. Ok for now though 😉

miteshvp:
Tests are failing. Need to revert this

jpopelka:
> Tests are failing. Need to revert this

Would be nice to have link(s) to the tests here as I'm having difficulties to find it now.

PR #493, opened 2018-01-10
Add a new  Storage Kronos Handler

sara-02:
For issue https://github.com/openshiftio/openshift.io/issues/1694


f8a_worker/dispatcher/nodes.yml
```diff
@@ -500,6 +500,15 @@
         encryption: false
         versioned: true
 
+    - name: 'S3KronosAppend'
+      import: 'f8a_worker.storages'
+      configuration:
+        <<: *configurationS3
+        bucket_name: 'dev-stack-analysis-clean-data'
```
msrb: Wouldn't be better to prefix the bucket with deployment prefix so different deployments don't share the same bucket? Take a look for example here: https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/493/files#diff-7847c95658ce3aa270db29af1f9a2aabR471

f8a_worker/storages/s3_kronos_append.py
```diff
@@ -0,0 +1,11 @@
+from . import AmazonS3
+
+
+class S3KronosAppend(AmazonS3):
+
+    def store_updated_data(self, input_json, file_name):
+        self.store_dict(input_json, file_name)
```
msrb: Here you're just passing args from one method to another. The same in `fetch_existing_data`. Not really a problem though.

msrb:
I think the bucket name should be prefixed so different deployment don't share the data. WDYT?

f8a_worker/dispatcher/nodes.yml
sara-02: Currently we don't have a prefix for this bucket.

f8a_worker/storages/s3_kronos_append.py
sara-02: Can we directly do that without creating a new storage class?

f8a_worker/dispatcher/nodes.yml
msrb: We will definitely need to have a separate bucket for different deployments in future. Since we want to merge this to master and possibly promote it to prod in the near future, wouldn't be better to prefix the bucket now?

sara-02: Ok, we can do that, and have the parameter take no value.

msrb: `bucket_name: '{DEPLOYMENT_PREFIX}-stack-analysis-clean-data'` would always point to a deployement-specific bucket. You don't have to worry about the DEPLOYMENT_PREFIX, it will be automatically substituted for the correct value. We already use this trick on other places.

msrb: And just to clarify: correct value means "prod" for prod, "STAGE" for stage, "<your-dev-cluster-username>" for your dev cluster deployment and so on.

f8a_worker/storages/s3_kronos_append.py
jpopelka: Yes, I'd say it should be possible to use `AmazonS3` and its `store_dict()` and `retrieve_dict()` directly.

The `node.yaml` entry should I think (per [doc](http://selinon.readthedocs.io/en/latest/yaml.html#storages)) look like:
```
- name: 'S3KronosAppend'
  import: 'f8a_worker.storages'
  classname: 'AmazonS3'
```

sara-02:
@msrb @jpopelka I have removed the extra storage handler file and appended the required info in the `node.yml` PTAL

PR #494, opened 2018-01-11
revert 'Reuse WORKER_RUN_DB_MIGRATIONS to configure SQS queues only from one pod.'

jpopelka:
When testing https://github.com/fabric8-analytics/fabric8-analytics-common/pull/446 I realized that only '_api_' queues are configured during deployment.

This PR reverts c671b4dfd748, which I think causes the problem.

The `worker-pre-hook.sh` change had been added due to the `sleep(1/5)` in `queue_conf.py`, but we actually don't need either, since I've tested that SQS is able to properly handle many tagging requests per second - I created a script which tries to tag queues in parallel processes without any sleeping and all queues were tagged correctly.

### DIFF
--- a/hack/queue_conf.py
+++ b/hack/queue_conf.py
@@ -61,10 +61,6 @@ def set_queue_attributes(queue_names):
             )
             _logger.debug("Remote responded with %r after setting queue tag for %r",
                           response, queue_name)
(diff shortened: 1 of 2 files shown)


jpopelka:
WIP, let me test this on `dev` first.

jpopelka:
[test]

jpopelka:
looks OK, ready for review

PR #495, opened 2018-01-15
Add synced2graph column to version table

tuxdna:
Required for - https://github.com/fabric8-analytics/fabric8-analytics-data-model/pull/69


tuxdna:
[test this]

tuxdna:
[test]

f8a_worker/models.py
```diff
@@ -138,6 +138,7 @@ class Version(Base):
     id = Column(Integer, primary_key=True)
     package_id = Column(Integer, ForeignKey(Package.id))
     identifier = Column(String(255), index=True)
+    synced2graph = Column(Boolean, nullable=False, default=False)
```
msrb: @tuxdna 

Question: who is going to set the flag to `False` when we update for example GitHub data for the package that was previously successfully synced to graph?

tuxdna: The default value is `False` so existing `INSERT` statements don't need to do anything. The first time an EPV is synced, **data-model-importer** will update the value to `True`. However you are right that, for cases where an existing EPV's data is update by some worker, we need to update this flag back to`False` again.

tuxdna: Let me add code set this flag on any update to EPV

tuxdna:
I think following should do the trick for an EPV:

```diff
diff --git a/f8a_worker/workers/result_collector.py b/f8a_worker/workers/result_collector.py
index 29bd0d0..7bc7ee2 100644
--- a/f8a_worker/workers/result_collector.py
+++ b/f8a_worker/workers/result_collector.py
@@ -25,6 +25,7 @@ class _ResultCollectorBase(BaseTask):
             # Substitute task's result with version that we got on S3
             worker_result.task_result = {'version_id': version_id}
 
+        results.version.synced2graph = True
         try:
             postgres.session.commit()
         except SQLAlchemyError:

```

@msrb Would you suggest otherwise?

tuxdna:
@msrb: Updated the result collect, which updates the flag once all workers finish.

tuxdna:
@msrb: Yeah, tested the migration together with all services using docker compose

msrb:
Let's test this in stage ;)

PR #497, opened 2018-01-19
Revert "Revert "Add feedback schema model""

msrb:
Reverts fabric8-analytics/fabric8-analytics-worker#496


miteshvp:
This should be merged now. I guess Server PR https://github.com/fabric8-analytics/fabric8-analytics-server/pull/236/files was causing the issue.

miteshvp:
Let's test it on Stage

msrb:
@miteshvp @sara-02 alembic actually remembers what's the actual version is (in a table in RDS) so reverting code changes is not enough unfortunately.

miteshvp:
Ok, that explains everything. Thanks. How do we go about it?

msrb:
@miteshvp if we can run a command in the worker container, then we can "downgrade" to previous version: http://alembic.zzzcomputing.com/en/latest/tutorial.html#downgrading

If we don't have access to the container, we unfortunately need to generate a new alembic migration on top of the current one, which will revert the changes.

PR #502, opened 2018-01-30
JSONB column for storing the dependency-snapshot

ravsa:
Additional JSONB column for storing the dependency-snapshot


miteshvp:
Lets test it on stage

PR #511, opened 2018-02-06
[LibrariesIoTask] move getting of project url to F8AConfiguration

jpopelka:
Just small refactoring.

### DIFF
--- a/f8a_worker/defaults.py
+++ b/f8a_worker/defaults.py
@@ -155,5 +155,26 @@ class F8AConfiguration(object):
             raise F8AConfigurationException("Limit for unauthorized GitHub access exceeded.")
         return token, headers
 
+    @classmethod
+    def libraries_io_project_url(cls, ecosystem, name):
+        """Construct url to endpoint, which gets information about a project and it's versions."""
+        url = '{api}/{platform}/{name}'. \
+            format(api=cls.LIBRARIES_IO_API,
+                   platform=ecosystem,
+                   name=name)
+
+        if not cls.LIBRARIES_IO_TOKEN or cls.LIBRARIES_IO_TOKEN == 'not-set':
+            raise F8AConfigurationException("LIBRARIES_IO_TOKEN has not been set.")
+
+        # 'no-token' value forces the API call to not use ANY token.
+        # It works, but if abused, they can ban your IP, so use with caution.
+        if cls.LIBRARIES_IO_TOKEN != 'no-token':
+            logger.warning("Libraries.io API calls will be without an API key."
+                           "It'll work, but if you're going to analyse more packages,"
+                           "please set the LIBRARIES_IO_TOKEN to your private token.")
+            url += '?api_key=' + cls.LIBRARIES_IO_TOKEN
+
+        return url
+
 
 configuration = F8AConfiguration()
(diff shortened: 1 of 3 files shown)


f8a_worker/defaults.py
```diff
@@ -155,5 +155,26 @@ def select_random_github_token(cls):
...
 
+    @classmethod
+    def libraries_io_project_url(cls, ecosystem, name):
+        """Construct url to endpoint, which gets information about a project and it's versions."""
+        url = '{api}/{platform}/{name}'. \
+            format(api=cls.LIBRARIES_IO_API,
+                   platform=ecosystem,
+                   name=name)
+
+        if not cls.LIBRARIES_IO_TOKEN or cls.LIBRARIES_IO_TOKEN == 'not-set':
```
pkajaba: What is the difference between `not-set` and and `no-token`?

f8a_worker/defaults.py
```diff
@@ -155,5 +155,26 @@ def select_random_github_token(cls):
...
+            raise F8AConfigurationException("LIBRARIES_IO_TOKEN has not been set.")
+
+        # 'no-token' value forces the API call to not use ANY token.
+        # It works, but if abused, they can ban your IP, so use with caution.
+        if cls.LIBRARIES_IO_TOKEN != 'no-token':
```
pkajaba: Is this condition really required? I believe that `not-set` and `no-token` are same, so this is redundant.

f8a_worker/workers/libraries_io.py
```diff
@@ -45,7 +29,8 @@ def execute(self, arguments):
         if ecosystem == 'go':
             name = quote(name, safe='')
 
-        project = get_response(self.project_url(ecosystem, name))
+        project_url = self.configuration.libraries_io_project_url(ecosystem, name)
```
pkajaba: Why did you move `libraries_io_project_url` to configuration module? It makes more sense to place it inside `utils` for me.

f8a_worker/defaults.py
jpopelka: `not-set` is default value, see
https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/f8a_worker/defaults.py#L58
https://github.com/fabric8-analytics/fabric8-analytics-deployment/blob/master/openshift/env-template.sh#L33
It means that user has not set the `LIBRARIES_IO_TOKEN` and we want to explicitly tell the user that it is required to do so.

`no-token` is special value. By setting `LIBRARIES_IO_TOKEN` to `no-token` the user requests that (s)he explicitly does not want to use ANY token, see the comment in code below. We use this in tests where we don't want to store the token anywhere, but still be able to run (and pass) the tests locally.
I understand that `not-set` and `no-token` sounds similar, so if you have any suggestions, how to call it, fire.
Maybe `unauthenticated` ?

f8a_worker/defaults.py
jpopelka: Yes, it is. See above comment.

f8a_worker/workers/libraries_io.py
jpopelka: Because it's basically a property of `F8AConfiguration`. I'd mark it `@property` if it had not arguments (ecosystem, name).

f8a_worker/defaults.py
pkajaba: No, it's OK. I just wanted to be sure whether it's not mistaken. Thanks!

f8a_worker/workers/libraries_io.py
pkajaba: Ok, thanks for an answer.

PR #514, opened 2018-02-08
Improve adding of new monitored upstream

jpopelka:
With this PR:
- We drop `active` flag from `monitored_upstreams` table, because we can check for the same with `.filter(Upstream.deactivated_at.is_(None))`
- When we're adding new monitored upstream record, we first check whether there is an older deactivated record with the same url. If it is, then we activate it, rather than adding completely new one (with the same url).

Fixes https://github.com/openshiftio/openshift.io/issues/2186


f8a_worker/workers/init_package_flow.py
```diff
@@ -41,30 +39,44 @@ def get_upstream_entry(self, db, package, url):
...
+            older_upstream = db.query(Upstream) \
+                .filter(Upstream.package_id == package.id) \
+                .filter(Upstream.deactivated_at.isnot(None)) \
+                .filter(Upstream.url == url) \
+                .order_by(desc(Upstream.updated_at)) \
+                .first()
+            if older_upstream:
+                self.log.info("Activating older upstream record entry for package %s/%s and "
+                              "upstream URL '%s'", package.ecosystem.name, package.name, url)
+                older_upstream.updated_at = now
```
msrb: Setting this here will break the following condition: https://github.com/jpopelka/fabric8-analytics-worker/blob/58a05e3d6e14c858acb5097cf83ee894f2f8af1f/f8a_worker/workers/init_package_flow.py#L100

f8a_worker/workers/init_package_flow.py
```diff
@@ -41,30 +39,44 @@ def get_upstream_entry(self, db, package, url):
...
+                older_upstream.deactivated_at = None
+                db.commit()
```
msrb: Not needed here (?)

f8a_worker/workers/init_package_flow.py
```diff
@@ -41,30 +39,44 @@ def get_upstream_entry(self, db, package, url):
...
         for entry in upstreams:
             if url is not None and entry.url != url:
                 self.log.info("Marking upstream entry with id '%s' and URL '%s' for deactivation, "
                               "substituting with upstream URL '%s'", entry.id, entry.url, url)
+                # deactivate entries that have different upstream URL
                 entry.deactivated_at = now
             else:
                 self.log.info("Reusing already existing and active upstream record with id '%d' "
                               "and upstream URL '%s'", entry.id, entry.url)
                 ret = entry
```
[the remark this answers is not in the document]
jpopelka: Good catch, I think it should be the newest. Fixed with 17382a6

f8a_worker/workers/init_package_flow.py
jpopelka: You're right. This line can be omitted since we set the `updated_at` later in `execute()`

PR #519, opened 2018-02-14
[CVEcheckerTask] Don't upload dep-check db if update fails

jpopelka:
`TimedCommand.get_command_output()` returns `[]` if there's a problem and `graceful` is `True` (which it is by default).


jpopelka:
[:-(](https://ci.centos.org/job/devtools-f8a-master-deploy-e2e-test/753/console)

PR #522, opened 2018-02-20
[CVEcheckerTask] Set JVM memory limit before running dependency-check

jpopelka:
Should fix https://github.com/openshiftio/openshift.io/issues/2222

### DIFF
--- a/f8a_worker/workers/CVEchecker.py
+++ b/f8a_worker/workers/CVEchecker.py
@@ -25,6 +25,8 @@ class CVEcheckerTask(BaseTask):
     _analysis_name = 'security_issues'
     schema_ref = SchemaRef(_analysis_name, '3-0-1')
 
+    dependency_check_jvm_mem_limit = '-Xmx768m'
+
     @staticmethod
     def get_cve_impact(id_):
         score = 0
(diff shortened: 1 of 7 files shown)


f8a_worker/workers/CVEchecker.py
```diff
@@ -153,23 +155,24 @@ def _npm_scan(self, arguments):
...
+        depcheck = configuration.dependency_check_script_path
         with tempdir() as temp_data_dir:
             s3.retrieve_depcheck_db_if_exists(temp_data_dir)
+            old_java_opts = os.getenv('JAVA_OPTS', '')
+            os.environ['JAVA_OPTS'] = CVEcheckerTask.dependency_check_jvm_mem_limit
             # give DependencyCheck 25 minutes to download the DB
             if TimedCommand.get_command_output([depcheck, '--updateonly', '--data', temp_data_dir],
                                                timeout=1500):
                 s3.store_depcheck_db(temp_data_dir)
+            os.environ['JAVA_OPTS'] = old_java_opts
```
msrb: What if the command throws an exception? Wouldn't be better to modify environment just for the subcommand? Like for example in https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/f8a_worker/workers/mercator.py#L205-L209.

jpopelka: Oh, this is what I was looking for but haven't found, thanks.
However the `TimedCommand.run()` works differently than `TimedCommand.get_command_output()`, so I'd rather stick with the `get_command_output()` if you don't mind :-)
It does not throw an exception, because it runs with `graceful=True`.

jpopelka:
[Build is failing](https://ci.centos.org/job/devtools-fabric8-analytics-worker-fabric8-analytics/533/console) due to dependencies conflicts. [New worker-base](https://github.com/fabric8-analytics/fabric8-analytics-worker-base/pull/31) should fix it.

jpopelka:
Looks OK when testing locally. Ready.

PR #524, opened 2018-02-21
Add index for synced2graph column

tuxdna:
Fixes issue: https://github.com/fabric8-analytics/fabric8-analytics-data-model/issues/100


tuxdna:
@msrb: That was quick!

tuxdna:
@msrb: Thanks!

PR #526, opened 2018-02-22
There was missing dollar sign in template which is causing errors

pkajaba:
It's not resolving Sentry DSN correctly.


pkajaba:
I will go ahead and merge for myself :octocat:

PR #528, opened 2018-02-22
Advanced python dependency solving

jpopelka:


msrb:
Oh, man... this is crazy... /me wondering if there is a field-tested library that would do this for us.

jpopelka:
> Oh, man... this is crazy...

yeah, podvody was writing code like crazy :-D

more simplification follows in #533

PR #530, opened 2018-02-23
[tests] Throw an exception on status code other than 200

msrb:
"Fixes": https://ci.centos.org/job/devtools-fabric8-analytics-worker-fabric8-analytics/545/consoleFull

The NIST website was down during the build.


tisnik:
@msrb, it looks good. Tests are passing -> LGTM

PR #531, opened 2018-02-23
[WIP][tests] Get rid of create_test_instance class method

msrb:


msrb:
Ouch, good catch. What a mess... let me open PRs for those first.

PR #534, opened 2018-02-26
[WIP] Add persistent volume and use it for Dependency-check DB update

jpopelka:
see https://github.com/openshiftio/openshift.io/issues/2222#issuecomment-367807721


msrb:
@jpopelka Ah, sorry, I wasn't very clear about this in my comment. We definitely don't want to add another PV just for DependencyCheck. We use the current PV as `/tmp/` with 10 gigs of space, i.e. we clean up when we are done with it. So we can reuse the same PV also for DependencyCheck, we just want to make sure that it's doing its business in its own directory.

msrb:
DependencyCheck is no moar.

PR #543, opened 2018-03-02
Test GithubDependencyTreeTask

jpopelka:
test for #536

### DIFF
--- a/tests/workers/test_github_dependency_tree.py
+++ b/tests/workers/test_github_dependency_tree.py
@@ -1,0 +1,29 @@
+"""Test GithubDependencyTreeTask."""
+
+import pytest
+
+from f8a_worker.workers import GithubDependencyTreeTask
+
+
+@pytest.mark.usefixtures("dispatcher_setup")
+class TestGithubDependencyTreeTask(object):
+    """Test GithubDependencyTreeTask."""
+
+    def test_maven_index_checker_repo(self):
+        """Test GithubDependencyTreeTask."""
+        args = {'github_repo': 'https://github.com/fabric8-analytics/maven-index-checker.git',
+                'github_sha': 'de64e1534724e53766dda472e9418350b85a1521',
+                'email_ids': 'dummy'}
+        task = GithubDependencyTreeTask.create_test_instance(task_name='dependency_tree')
+        results = task.execute(args)
+
+        assert isinstance(results, dict)
+        assert set(results.keys()) == {'dependencies'}
+        obtained_dependencies = set(results['dependencies'])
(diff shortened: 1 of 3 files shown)


tests/workers/test_github_dependency_tree.py
```diff
@@ -0,0 +1,28 @@
...
+
+        assert isinstance(results, dict)
+        assert set(results.keys()) == {'dependencies'}
+        expected_direct_dependencies = ['com.googlecode.json-simple:json-simple:1.1.1',
+                                        'org.apache.maven.indexer:indexer-core:6.0.0']
+        expected_transitive_dependencies = ['commons-io:commons-io:2.0.1',
+                                            'com.google.guava:guava:20.0',
+                                            'junit:junit:4.10']
+        for dep in expected_direct_dependencies + expected_transitive_dependencies:
+            assert dep in results['dependencies']
```
abs51295: Storing the results in a `set` would result in constant time lookup. Maybe use something like `obtained_dependencies = set(results.get('dependencies', []))`?

tests/workers/test_github_dependency_tree.py
```diff
@@ -0,0 +1,29 @@
...
+        assert isinstance(results, dict)
+        assert set(results.keys()) == {'dependencies'}
+        obtained_dependencies = set(results['dependencies'])
+        expected_direct_dependencies = {'com.googlecode.json-simple:json-simple:1.1.1',
+                                        'org.apache.maven.indexer:indexer-core:6.0.0'}
+        assert expected_direct_dependencies.issubset(obtained_dependencies)
+        expected_transitive_dependencies = {'commons-io:commons-io:2.0.1',
+                                            'com.google.guava:guava:20.0',
+                                            'junit:junit:4.10'}
+        assert expected_transitive_dependencies.issubset(obtained_dependencies)
```
abs51295: I think obtained dependencies should be the exact same as `expected_direct_dependencies + expected_transitive_dependencies`. It can look something like this:

```Python
expected_dependencies = expected_direct_dependencies.union(expected_transitive_dependencies)
assertSetEqual(obtained_dependencies, expected_dependencies)
```

jpopelka: We don't need to test whether `maven-dependency-plugin` works correctly,
we just need to know whether the task runs it and the test as it is answers that sufficiently.

abs51295:
@jpopelka I have approved the changes. If you can resolve merge conflicts, I can merge it as well.

tisnik:
@jpopelka good to know that the new CI job works as expected ;)

PR #544, opened 2018-03-02
Add missing docstrings

jpopelka:


f8a_worker/workers/schemas/__init__.py
```diff
@@ -0,0 +1 @@
+"""Obviously empty file dude."""
```
tisnik: cool :)

tisnik:
Nice @jpopelka . TYVM. Actually I was started working on the same issue, so let's merge this one not to do the same thing again :)

PR #547, opened 2018-03-05
Store available results on S3, regardless of failure

msrb:
The results from tasks that finished successfully were never stored on S3 if there was an error later on.


jpopelka:
[test]

PR #548, opened 2018-03-06
Small clean up

msrb:
Don't install `scipy` (almost 200MiB) if we don't need it.


msrb:
Weird...

msrb:
scikit-learn requires scipy, but doesn't pull it in automatically :-1: :hankey:

jpopelka:
Then maybe you could have added a comment into the `requirements.in` about it ?

msrb:
> Then maybe you could have added a comment into the requirements.in about it ?

Good idea: https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/550

PR #551, opened 2018-03-06
[alembic] Keep just one head

msrb:
Fixes:
"Multiple head revisions are present for given argument 'head'"


msrb:
Testing this now...

PR #567, opened 2018-03-13
[Git.clone()] recreate dir on error if previously existed

jpopelka:
`git clone repo dir/` deletes `dir/` if cloning fails. There's no option to turn this off.
This might confuse caller of the method, who could want to delete the dir itself.
Therefore we recreate the dir/ on error.

Example where this happens is [GitStats._get_log()](https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/f8a_worker/workers/git_stats.py#L28)
See https://errortracking.prod-preview.openshift.io/openshift_io/fabric8-analytics-production/issues/705/
which fails with cryptic `No such file or directory: '/tmp/tmpkllc_dvr'` instead of self-explaining `'Unable to clone: https://github.com/babel-plugins/babel-plugin-langpack'`

### DIFF
--- a/f8a_worker/process.py
+++ b/f8a_worker/process.py
@@ -8,7 +8,7 @@ import requests
 
 from git2json.parser import parse_commits
 from git2json import run_git_log
-
+from pathlib import Path
 from re import compile as re_compile
 from urllib.parse import urljoin, urlparse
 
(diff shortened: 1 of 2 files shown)


f8a_worker/process.py
```diff
@@ -53,21 +53,35 @@ def clone(cls, url, path, timeout=300, depth=None, branch=None, single_branch=Fa
...
+            cls.config()
             TimedCommand.get_command_output(cmd, graceful=False, timeout=timeout)
         except TaskError as exc:
+            if not path.is_dir() and mode:
+                # 'git clone repo dir/' deletes (no way to turn this off) dir/ if cloning fails.
+                # This might confuse caller of this method, so we recreate the dir on error here.
+                try:
+                    path.mkdir(mode)
+                except OSError:
+                    pass
```
pkajaba: I don't like `pass` here. At least some logging would be much nicer.

PR #576, opened 2018-03-16
yaml validator

jpopelka:


tisnik:
LGTM. The pyyaml library is in requirements.txt so feel free to fix^Wmerge it.

PR #577, opened 2018-03-16
Remove nowait attribute from GraphImporterTask, PackageGraphImporterTask and bayesianPackageFlow

tuxdna:
Fixes: https://github.com/openshiftio/openshift.io/issues/2219

### DIFF
--- a/f8a_worker/dispatcher/flows/bayesianAnalysisFlow.yml
+++ b/f8a_worker/dispatcher/flows/bayesianAnalysisFlow.yml
@@ -8,7 +8,7 @@
        - 'bayesianFlow'
        - 'bayesianApiFlow'
        - 'bayesianPriorityFlow'
-       - 'bayesianPackageFlow'
+       # - 'bayesianPackageFlow'
        - 'bayesianApiPackageFlow'
        - 'bayesianPriorityPackageFlow'
       propagate_parent:
--- a/f8a_worker/dispatcher/flows/bayesianFlow.yml
+++ b/f8a_worker/dispatcher/flows/bayesianFlow.yml
@@ -7,8 +7,8 @@
       propagate_node_args: true
       # propagate arguments from InitAnalysisFlow to bayesianAnalysisFlow
       node_args_from_first: true
-      nowait:
-        - 'GraphImporterTask'
+#      nowait:
+#        - 'GraphImporterTask'
       sampling:
         name: 'constant'
         args:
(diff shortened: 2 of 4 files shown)


msrb:
@tuxdna this PR will not fix the no-such-bucket error, is that correct?

We don't wait for graph importer to finish as it was (still is?) quite slow operation and we don't want to block the pipeline, if the only thing we do is that we synchronously wait for graph ingestion to finish. Graph importer task is also running on dedicated workers as we don't want all workers to wait for graph.

tuxdna:
@msrb: This PR does fix the no-such-bucket error.

From `nowait` in docs here - https://selinon.readthedocs.io/en/latest/yaml.html#flow-definition

`nowait` is not same as **synchronous**. In the absence of `nowait` the individual nodes do run asynchronously. However the semantic is to

> wait for all nodes to complete in order to end flow

I hope that makes this changes in this PR clearer.

msrb:
@tuxdna that doesn't sound right. We call graph importer after result collector: https://github.com/tuxdna/fabric8-analytics-worker/blob/bc40203fc0a6761f43e9b788b25b2ac1274d8c79/f8a_worker/dispatcher/flows/bayesianFlow.yml#L54-L55

Which means that when we call graph importer, the data is already on S3. We then just don't wait for graph import to finish and we end the flow (graph import is the last step anyway, so it's fine not to wait).

But maybe I am missing something here. Let me check locally.

tuxdna:
At the moment Graph importer is not as slow as it used to be earlier. We have much better write performance, so slowness shouldn't be an issue.

Yes please check locally. You may want to observe the overall time it takes to run different workers vs. the time it takes for actual graph import for an EPV.

f8a_worker/dispatcher/flows/bayesianAnalysisFlow.yml
```diff
@@ -8,7 +8,7 @@
        - 'bayesianFlow'
        - 'bayesianApiFlow'
        - 'bayesianPriorityFlow'
-       - 'bayesianPackageFlow'
+       # - 'bayesianPackageFlow'
```
msrb: Ah, I see now. I see why the no-such-bucket error disappeared. It's because of this change. We will wait for package level data to land in S3/minio before we call GraphImporter task. And we call the task twice, once for component-level data and second time for package-level data.

To be honest, this feels like a workaround. With this change we will keep the flow control message on `bayesianAnalysisFlow` queue much longer than actually needed. Dispatcher workers will need to periodically check the state of the flow. This could slow the system down as we talk about thousands (even 50k-100k sometimes) of messages which will consume some of our precious resources.

@tuxdna let's talk this through :wink:

tuxdna: Update the discussion here: https://github.com/openshiftio/openshift.io/issues/2219#issuecomment-375276749

PR #578, opened 2018-03-19
Swap date difference, because it is always negative, which is resulti…

pkajaba:
…ng to True in any case


pkajaba:
This issue was actually caused by 2 problems. First is that condition which is fixed in this PR always returned `True`, which restricted further updates to Package Level data.

Second is that we had to have some inconsistency in our PostgreSQL database, because those analysis have to be missing even though there is upstream URL registered. 

@msrb is it possible that those results were deleted in past? I don't have access to prod DB, but it acted the same way when I deleted worker results by hand. 

I have an question/suggestion - do we even need this check whether there was package level analysis performed recently?

PR #580, opened 2018-03-20
Adding support for sentry

GeetikaBatra:


tuxdna:
What steps are required to check that notifications are sent to Sentry ?

GeetikaBatra:
@tuxdna, I took reference from here, https://github.com/getsentry/sentry/blob/96bc1c63df5ec73fe12c136ada11561bf52f1ec9/src/sentry/celery.py
If this is the right way, in your opinion, we can merge it so that it is available on staging and watch for logs on Red Hat sentry.

f8a_worker/start.py
```diff
@@ -2,8 +2,21 @@
...
 """Start the application."""
 
-from celery import Celery
+import celery
+import os
 from f8a_worker.setup_celery import init_celery, init_selinon
+from raven.contrib.celery import register_signal, register_logger_signal
+
+
+class Celery(celery.Celery):
```
tuxdna: You may want to name this class something like `SentryCelery` insead of `Celery`.

f8a_worker/start.py
```diff
@@ -2,8 +2,21 @@
...
+
+    def on_configure(self):
+        """Set up sentry client."""
+        dsn = os.environ.get("SENTRY_DSN")
+        client = raven.Client(dsn)
+        register_logger_signal(client)
+        register_signal(client)
 
 
 app = Celery('tasks')
```
tuxdna: .. and use the new class name here.

GeetikaBatra:
@tuxdna, done. :)

tuxdna:
LGTM.

pkajaba:
@GeetikaBatra what do you want to achieve here?

GeetikaBatra:
@pkajaba, I am trying to address this issue here https://github.com/openshiftio/openshift.io/issues/2535
Currently, we do not catch exceptions for worker in sentry. This patch is an iteration to check if sentry logs issue for worker running on staging.

pkajaba:
@GeetikaBatra great, and have you tried it whether it does not send events twice?

GeetikaBatra:
@pkajaba,  I had a discussion with the team and everyone suggested that we can observe if the errors are logged on staging directly as it would be difficult to reproduce this in local. @msrb can shed some more light on it. :)

PR #588, opened 2018-03-22
Refactored init task for package flow

pkajaba:


f8a_worker/dispatcher/flows/bayesianPackageFlow.yml
```diff
@@ -15,6 +15,8 @@
       edges:
         - from:
           to: 'InitPackageFlow'
+        - from: 'bayesianPackageAnalysisFlow'
+          to: 'UpstreamUpdateTask'
```
msrb: Does this mean that upstream update task and the flow task will run simultaneously?

f8a_worker/workers/init_package_flow.py
```diff
@@ -115,36 +38,21 @@ def execute(self, arguments):
...
-                    and datetime.datetime.utcnow() - upstream.updated_at < self._UPDATE_INTERVAL:
-                self.log.info('Skipping upstream package check as data are considered as recent - '
-                              'last update %s.',
-                              upstream.updated_at)
-                # keep track of start, but do not schedule nothing more
-                # discard changes like updates
-                db.rollback()
+        if not arguments.get('force') and package_analysis is not None:
+            if datetime.datetime.utcnow() - package_analysis.started_at < self._UPDATE_INTERVAL:
+                InitPackageFlow.inject_arguments(arguments, package_analysis.id, package.id)
```
msrb: Injecting `document_id` argument beats the purpose of this is-older-than condition. Presence of the argument later controls whether we should run github task and others: https://github.com/pkajaba/fabric8-analytics-worker/blob/df19a286b553b8a41460a6daf45f79d0a29a0e17/f8a_worker/dispatcher/flows/bayesianPackageFlow.yml#L23-L25. So if we care about how old the previous analysis is, then we cannot inject the `document_id` argument in this case.

f8a_worker/dispatcher/flows/bayesianPackageFlow.yml
pkajaba: Well, it wasn't my intention. I wanted to run it one after another. Does it mean that current configuration is bad?

f8a_worker/workers/init_package_flow.py
```diff
@@ -115,36 +38,21 @@ def execute(self, arguments):
         except NoResultFound:
             raise FatalTaskError('Unknown ecosystem: %r' % arguments['ecosystem'])
         package = Package.get_or_create(db, ecosystem_id=ecosystem.id, name=arguments['name'])
-        url = self.get_upstream_url(arguments)
-        upstream = self.get_upstream_entry(package, url)
-        if upstream is None:
-            upstream = self.add_or_update_upstream(package, url)
-        arguments['url'] = upstream.url
```
msrb: I think not having the URL in arguments could cause problems down the road. For example in GitHub task: https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/08f5af72f7f24105541659f103e0e271c6e2d3d2/f8a_worker/workers/githuber.py#L82

msrb:
Big +1 on refactoring, this part needs it very badly. But the current configuration and code is so complicated that I think it would be probably better to do it separately. Let's open an issue for refactoring. I have an idea how to fix the bug. Will open a PR in a minute.

msrb:
@pkajaba PTAL: https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/603

humaton:
We don't have the capacity to continue on this one. Closing.

PR #594, opened 2018-03-27
Only rely on VictimsDB for Maven packages

msrb:
No false positives from now on.

- skip DependencyCheck
- exclude Maven from CVEDB sync
- mark vulnerable dependencies directly in graph, if possible
- make sure we have all verions of vulnerable packages in our system


pkajaba:
[test]

msrb:
Rebased to pull in CVE test fixes from Pavel. And corrected one test that we are adding in this PR.

f8a_worker/dispatcher/flows/victimsFlow.yml
```diff
@@ -0,0 +1,11 @@
+---
+  flow-definitions:
+    - name: 'victimsFlow'
+      queue: '{DEPLOYMENT_PREFIX}_{WORKER_ADMINISTRATION_REGION}_victimsFlow_v0'
+      sampling:
+        name: 'constant'
+        args:
+          retry: 20
+      edges:
+        - from:
```
pkajaba: So this flow will be started just from jobs service right?

msrb: My idea is to trigger it (via jobs) when we push something to master branch.

msrb:
@pkajaba Thanks. I don't like the way how it is now :) We don't need to analyze packages to discover that they are vulnerable. We get CVE data from other sources, not by analyzing. But currently the problem is that literally everything in fabric8-analytics expects packages to be analyzed first.

I think we should distinguish between package/version discovery - when we found out about new package or version and actual analysis. I.e. we could have all discovered versions in ingestion DB and in graph, even if we don't have any additional data for them. That way we could simply mark vulnerable packages in graph without need to analyze packages first. Not just that but it would once for all solved the problems with unknown "latest versions". And we would actually know how we are doing, how much of the ecosystem we already have analyzed and what's missing. We could also utilize resources better as there would be constantly packages that we could analyze, without people manually scheduling analyses via jobs service.

That's something we will hopefully start working on next sprint.

PR #598, opened 2018-04-06
Use get() instead of pop() to get the dependencies

abs51295:
- Fixes the problem of getting empty dependencies because they are not popped anymore!

Signed-off-by: abs51295 <aagams68@gmail.com>


abs51295:
@pkajaba Thanks for merging this PR, but I was curious as to why `pop()` was used. Am I missing some data normalization here?

pkajaba:
@abs51295 I don't know :-(

abs51295:
@msrb can you add something here?

PR #601, opened 2018-04-10
Fix go-get bug

msrb:


msrb:
Tests fail because of new pydocstyle: https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/602

runtest.sh
```diff
@@ -96,6 +96,7 @@ echo "Starting test suite"
 docker run -t \
   -v "${here}:/f8a_worker:ro,Z" \
   --network ${DOCKER_NETWORK} \
+  -u 9007 \
```
abs51295: Any reason why this was added and the chunk from `Dockerfile.tests` was removed?

msrb: The code runs under random non-existent user in OpenShift. We were explicitly creating a user in `Dockerfile.tests`. That's why the tests didn't catch this bug when it was introduced. With this change we are running the tests as a user with ID 9007, but there is no user with such ID in `/etc/passwd`.

abs51295: Makes sense. Thanks for the explanation 👍

abs51295:
[test]

PR #602, opened 2018-04-10
Pydocstyle 2.4 fixes

msrb:


msrb:
pydocstyle 2.4.0 was released today, and it's complaining more than ever.

https://pypi.python.org/pypi/pycodestyle

tisnik:
@msrb LGTM! Thanks a lot!

PR #605, opened 2018-04-13
moved this step to fabric8-analytics-worker-base

syamgk:
https://github.com/fabric8-analytics/fabric8-analytics-worker-base/commit/7365c673d4fe4cb2864a8d1f4475cb345f7f4eb7


msrb:
Now I realized that we will need to update the reference to latest -base image in Dockerfile. I will do that once the image is available.

msrb:
And since there are some problems with webhooks not reaching CI, I am going to revert this for now.

PR #607, opened 2018-04-16
Fix another integrity error issue in package level flows

msrb:
`document_id` has a very special meaning in package level flows,
so make sure we don't run around with the one inherited from parent flow.


msrb:
Another instance of https://github.com/openshiftio/openshift.io/issues/2218.

msrb:
[test]

msrb:
[test]

PR #609, opened 2018-04-18
Add support for Python for GithubDependencyTreeTask

abs51295:
- Uses mercator to get a list of direct (and indirect) dependencies by parsing requirements.txt file.


f8a_worker/workers/dependency_parser.py
```diff
@@ -144,3 +154,28 @@ def get_npm_dependencies(cls, path):
...
+        :param path: path to run the mercator
+        :return: set of direct (and indirect) dependencies
+        """
+        mercator_output = cls._mercator.run_mercator(arguments={"ecosystem": "pypi"},
+                                                     cache_path=path,
+                                                     resolve_poms=False)
+
+        set_package_names = set()
+        mercator_output_details = mercator_output['details'][0]
+        dependencies = mercator_output_details.get('dependencies')
```
samuzzal-choudhury: Set the default of `mercator_output_details .get` to an empty array.

samuzzal-choudhury: Rest LGTM.

abs51295: I don't think we should continue silently if we don't get the dependencies from mercator, since that would be a bug from mercator side.

msrb: I'd hope that mercator returns non-zero status code if it failed from some reason. Thus it should fail before we reach this point. Now the question is: will the `dependencies` key exist if `requirements.txt` file is empty? :)

abs51295: I think it will: 

```
{
  "items": [
    {
      "digests": {
        "manifest": "da39a3ee5e6b4b0d3255bfef95601890afd80709"
      },
      "ecosystem": "Python-RequirementsTXT",
      "path": "/workdir/requirements.txt",
      "result": {
        "dependencies": [],
        "ecosystem": "python-requirementstxt"
      },
      "time": "2018-04-25T17:07:38.0057701Z"
    }
  ],
  "versions": {
    "Python-RequirementsTXT": "c381b89f575af1ea6fd787a60e05c3a24f268b1e"
  }
}
```

samuzzal-choudhury: @abs51295 an additional check as you do not want your code to break in any case.

abs51295:
[test]

PR #611, opened 2018-04-19
Drop Jenkinsfile

msrb:


msrb:
The real purpose of this PR is to trigger CI :)

PR #612, opened 2018-04-19
Add support for package-lock.json file for GithubDependencyTreeTask

abs51295:
- Currently, mercator only supports npm-shrinkwrap.json and hence we need a hack to support package-lock.json as of now.
- Need to revert this once mercator supports package-lock.json.

Signed-off-by: abs51295 <aagams68@gmail.com>

### DIFF
--- a/f8a_worker/workers/dependency_parser.py
+++ b/f8a_worker/workers/dependency_parser.py
@@ -42,6 +42,10 @@ class GithubDependencyTreeTask(BaseTask):
             repo.reset(revision=github_sha, hard=True)
             with cwd(repo.repo_path):
                 # TODO: Make this task also work for files not present in root directory.
+
+                # First change the package-lock.json to npm-shrinkwrap.json
+                GithubDependencyTreeTask.change_package_lock_to_shrinkwrap()
+
                 if peek(Path.cwd().glob("pom.xml")):
                     return GithubDependencyTreeTask.get_maven_dependencies()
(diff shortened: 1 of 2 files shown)


abs51295:
Working on adding tests :)

abs51295:
Please merge #615 first :)

msrb:
[test]

f8a_worker/workers/dependency_parser.py
```diff
@@ -144,3 +148,17 @@ def get_npm_dependencies(cls, path):
...
+    @staticmethod
+    def change_package_lock_to_shrinkwrap():
+        """Rename package-lock.json to npm-shrinkwrap.json.
+
+        For more information about package-lock.json please visit
+        https://docs.npmjs.com/files/package-lock.json
+        """
+        # TODO: Remove this method once mercator has support for package-lock.json
+
+        package_lock_path = Path.cwd() / "package-lock.json"
```
msrb: Does this reliably work? I mean, can we expect CWD to be correct when we hit this line?

msrb:
@abs51295 just a question, otherwise LGTM.

f8a_worker/workers/dependency_parser.py
abs51295: Yes, because we are in correct directory in line 43 itself.

msrb: I see it now, thanks :wink:

PR #615, opened 2018-04-24
Fix the foolish mistake of not considering multiple transitive dependencies for NPM

abs51295:


abs51295:
[test]

f8a_worker/workers/dependency_parser.py
```diff
@@ -127,14 +127,15 @@ def get_npm_dependencies(cls, path):
                                           package=name, version=version))
 
                 if transitive_deps:
```
samuzzal-choudhury: This check is redundant as the following `for` statement handles the existence of the array.

abs51295: It will not, since we get `None` if transitive dependencies are not present and if you directly iterate a `None` object you will get `TypeError: 'NoneType' object is not iterable`

abs51295: But, we can let the `get` return a default empty list and then let for loop take care of it.

samuzzal-choudhury: @abs51295 The above is a cleaner approach.

PR #617, opened 2018-04-25
Migrate to pip 10

msrb:
Using Pip's internal API is exactly what we SHOULDN'T be doing, but we also don't want to reimplement the logic on our own. We want to parse requirement.txt files exactly as the native tool, Pip, would parse them.

We will see how this goes.


msrb:
Test failures are due to NIST website being unavailable at the moment:

```
OSError: Unable to reach URL: https://nvd.nist.gov/vuln/detail/2015-7501
```

msrb:
[test]

abs51295:
[test]

PR #620, opened 2018-04-26
Add support for Go ecosystem for GithubDependencyTreeTask

abs51295:
- Parse Gopkg.lock file via anymarkup to gather all direct and transitive dependencies along with their versions.
- Parse glide.lock file via mercator to gather all direct and transitive dependencies along with their versions.


abs51295:
Reopening a new one, messed the history I suppose

PR #622, opened 2018-04-26
Add go support for GithubDependencyTreeTask

abs51295:
- Parse Gopkg.lock file via anymarkup to gather all direct and transitive dependencies along with their versions.
- Parse glide.lock file via mercator to gather all direct and transitive dependencies along with their versions.
- Add a failover mechanism for osioAnalysisFlow so that ReportGenerationTask is called irrespective of failures.


abs51295:
I also added fail over mechanism in this PR, since this was a change that I later noticed after pushing the changes. Would have liked it in a separate PR but anyways.

PR #627, opened 2018-05-02
Bump worker-base image to f8a-worker-base:0bcf081 for rhel

tuxdna:
For: https://github.com/openshiftio/openshift.io/issues/3182


tuxdna:
Build passed, merging now.

PR #632, opened 2018-05-04
Raise an exception if IndianaJones try to fetch npm package with version

humaton:
Raise an exception if IndianaJones try to fetch npm package with version

https://github.com/openshiftio/openshift.io/issues/3405


f8a_worker/process.py
```diff
@@ -356,8 +356,14 @@ def fetch_npm_artifact(name, version, target_dir):
         #      └── package.tgz
         # 3 directories, 6 files
         name_ver = name
+        invalidChars = set(string.punctuation.replace(".", ""))
```
msrb: Aren't dashes also valid version characters in NPM? Like in `1.1.1-alpha1`. This is exactly my concern with this approach. We may easily refuse to analyze perfectly valid versions.

Idea: wouldn't be better to fetch list of all versions for the package in question and just check if the version given to us by user exists?

msrb:
@humaton added a comment, please check.

f8a_worker/process.py
humaton: Ok looking into sem ver definition only "." and "-" special characters are allowed.

f8a_worker/process.py
```diff
@@ -356,8 +356,14 @@ def fetch_npm_artifact(name, version, target_dir):
         #      └── package.tgz
         # 3 directories, 6 files
         name_ver = name
+        invalidChars = set(string.punctuation.replace(".", "").replace("-", ""))
```
msrb: I am not sure if this is going to work. Take a look here: https://docs.npmjs.com/misc/semver#usage

I can see following there:
`5.0.0 - 7.2.3`

None of the characters there is invalid for us, but yet it's a version range and not a specific version.

PR #635, opened 2018-05-04
Wait for graph in API workers

msrb:


msrb:
cc @samuzzal-choudhury @abs51295

PR #638, opened 2018-05-08
Fix Value error for engines in _handle_javascript data.

GeetikaBatra:
Ref Bug - https://github.com/openshiftio/openshift.io/issues/3406

Signed-off-by: Geetika Batra gbatra@redhat.com

### DIFF
--- a/f8a_worker/data_normalizer.py
+++ b/f8a_worker/data_normalizer.py
@@ -124,7 +124,6 @@ class DataNormalizer(object):
             return list(filter(None, iterable))
 
         base = self.transform_keys(data, key_map)
-
         # {'url': 'https://github.com/o/p/issues',
         #  'email': 'project@name.com'} -> 'https://github.com/o/p/issues <project@name.com>'
         if isinstance(base.get('bug_reporting'), dict):
@@ -227,8 +226,13 @@ class DataNormalizer(object):
                         base['engines'][name] = operator_version
         elif isinstance(engines, str):
             # 'node 4.2.3' -> {"node": "4.2.3"}
-            name, version = engines.split()
-            base['engines'] = {name: version}
+            try:
+                name, version = engines.split()
+                base['engines'] = {name: version}
+            except ValueError:
+                # For malformed data: '8.3' -> {}
+                base['engines'] = {}
+
         if base['engines'] is not None:
             for name, version_spec in base['engines'].items():
                 if ' ' in version_spec:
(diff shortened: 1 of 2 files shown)


f8a_worker/data_normalizer.py
```diff
@@ -227,8 +226,12 @@ def _rf(iterable):
                         base['engines'][name] = operator_version
         elif isinstance(engines, str):
             # 'node 4.2.3' -> {"node": "4.2.3"}
-            name, version = engines.split()
-            base['engines'] = {name: version}
+            try:
+                name, version = engines.split()
+                base['engines'] = {name: version}
+            except Exception:
```
msrb: We can be more specific here, i.e. we can catch `ValueError` exceptions.

msrb: Also I'd add comment here that we are giving up as the data is malformed.

jpopelka: Or just change `elif isinstance(engines, str):` to `elif isinstance(engines, str) and ' ' in engines:`.
Basically all the `_handle_javascript()` code is handling malformed data since nodejs package maintainers don't care about https://docs.npmjs.com/files/package.json and you'd soon end up with lots of `try` `expect`.

msrb: And that's why Node developers have no friends! :)

`elif isinstance(engines, str) and ' ' in engines:` would still throw an exception if there is more values to unpack. I'd maybe keep this simple and just catch the exception here. Exception means the data is weird and we don't care, we ignore it.

@jpopelka glad to see you here! :)

f8a_worker/data_normalizer.py
```diff
@@ -227,8 +226,13 @@ def _rf(iterable):
                         base['engines'][name] = operator_version
         elif isinstance(engines, str):
             # 'node 4.2.3' -> {"node": "4.2.3"}
-            name, version = engines.split()
-            base['engines'] = {name: version}
+            try:
+                name, version = engines.split()
+                base['engines'] = {name: version}
+            except ValueError:
```
GeetikaBatra: @msrb. Please take a look :)

f8a_worker/data_normalizer.py
```diff
@@ -227,8 +226,13 @@ def _rf(iterable):
...
+                # For malinformed data: '8.3' -> {}
```
msrb: Typo? :) https://www.urbandictionary.com/define.php?term=malinformed

f8a_worker/data_normalizer.py
```diff
@@ -227,8 +226,13 @@ def _rf(iterable):
                         base['engines'][name] = operator_version
         elif isinstance(engines, str):
             # 'node 4.2.3' -> {"node": "4.2.3"}
-            name, version = engines.split()
-            base['engines'] = {name: version}
+            try:
+                name, version = engines.split()
+                base['engines'] = {name: version}
+            except ValueError:
+                # For malformed data: '8.3' -> {}
```
GeetikaBatra: @msrb, PTAL

GeetikaBatra:
Thanks @msrb  :)

PR #640, opened 2018-05-14
[Openshift] Get SENTRY_DSN from secrets vault

pkajaba:


pkajaba:
I am not 100% whether this part can stay in template: https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/openshift/template.yaml#L247

@msrb can you tell whether it's ok or not?

msrb:
@pkajaba `SENTRY_DSN` parameter could stay in the template, but since we are no longer referencing it, then removing is probably a good idea.

PR #654, opened 2018-05-25
Remove StackAggregatorTasks

pkajaba:


f8a_worker/dispatcher/flows/componentApiFlow.yml
```diff
@@ -1,8 +0,0 @@
----
-  flow-definitions:
-    - name: 'componentApiFlow'
-      queue: '{DEPLOYMENT_PREFIX}_api_componentApiFlow_v0'
-      edges:
-        - from:
-          to:
-            - 'BookkeeperTask'
```
samuzzal-choudhury: @pkajaba We have got a dependency on componentApiFlow in the bayesian coreapi server [[1]](https://github.com/fabric8-analytics/fabric8-analytics-server/blob/master/bayesian/api_v1.py#L307) [[2]](https://github.com/fabric8-analytics/fabric8-analytics-server/blob/master/bayesian/utils.py#L81-L92) and that is the only way we can track the component analysis requests to the system. This will enable us to do user actions analytics in the future. So, let's not remove the flow and the task.

f8a_worker/dispatcher/nodes.yml
```diff
@@ -203,11 +174,6 @@
       selective_run_function:
         name: 'selective_run_function'
         import: 'f8a_worker.dispatcher.selective'
-    - name: 'BookkeeperTask'
```
samuzzal-choudhury: @pkajaba  let's not remove the BookkeeperTask because of the reason specified in the earlier comment.

f8a_worker/workers/__init__.py
```diff
@@ -1,7 +1,6 @@
 """Import all tasks."""
 
 from f8a_worker.workers.CVEchecker import CVEcheckerTask
-from f8a_worker.workers.bookkeeper import BookkeeperTask
```
samuzzal-choudhury: @pkajaba please restore the bookkeepertask.

f8a_worker/workers/bookkeeper.py
```diff
@@ -1,81 +0,0 @@
-"""Bookkeeper Task."""
```
samuzzal-choudhury: @pkajaba this file has to be restored.

pkajaba:
@samuzzal-choudhury changes were made.

PR #656, opened 2018-05-30
Fix ImportError in models.py

pkajaba:


pkajaba:
@msrb there was another occurrence of this import error.

PR #659, opened 2018-05-31
Fix version ranges support in Victims worker

msrb:
Fixes: https://github.com/openshiftio/openshift.io/issues/3645


f8a_worker/victims.py
```diff
@@ -251,7 +251,7 @@ def is_version_affected(affected_versions, checked_version):
                 else:
                     if checked_version <= version:
                         return True
-            return False
+        return False
```
msrb: omg... this was really stupid :)

PR #668, opened 2018-06-12
Ignore certain errors in Sentry (PR #1)

msrb:


msrb:
This is just a test. Will see how it goes.

PR #669, opened 2018-06-12
Do not raise NotImplementedError in store_error()

msrb:


msrb:
@tisnik Merging to test this on staging :wink: PTAL.

PR #684, opened 2018-06-28
Utilize FilteredVictimsDB in VictimsCheck task

msrb:


miteshvp:
@msrb IIUC, this will run either filteredflow or normalflow but not both?

msrb:
@miteshvp exactly. the purpose of filtered flow is to only report on certain subset of CVEs from victims DB (for example only those that were added/updated since last run). Normal flow would iterate over all CVEs every time - which would be a problem for gemini as it is only interested in new additions.

PR #686, opened 2018-07-02
Skip tagging related tasks

msrb:


msrb:
It is possible that E2E tests will fail for these changes.

msrb:
Let's see what E2E tests say...

PR #697, opened 2018-07-19
Run dependency snapshot task only conditionally

msrb:


msrb:
The results of the task are only useful if we want to recursively analyse the whole stack. They are not useful if we run analysis with `recursive_limit=0`.

msrb:
[test]

PR #702, opened 2018-07-30
Increase possible package name length

msrb:


msrb:
Package names in Golang are URLs, and those can be pretty long.

PR #712, opened 2018-08-28
Ignore value error while parsing output of mvn dependency:resolve

abs51295:


f8a_worker/workers/dependency_parser.py
```diff
@@ -156,7 +156,13 @@ def parse_maven_dependency_resolve(file):
                 if package_name:
                     # Remove scope from package name
                     package_name = package_name.rsplit(':', 1)[0]
-                    add_maven_coords_to_set(package_name, set_package_names)
+                    try:
+                        add_maven_coords_to_set(package_name, set_package_names)
+                    except ValueError:
+                        # We expect some value errors here because the file might contain
+                        # plain english sentences. For example:
```
msrb: :) I think it would be better to do some preprocessing on client side and send the list of dependencies in some predefined format.

Is it OK to ignore the exception here completely? No logging needed?

msrb:
@abs51295 just a question, PTAL :wink:

f8a_worker/workers/dependency_parser.py
abs51295: Well, the command runs inside the container and hence there can be no pre-processing done. I considered logging, but this is a static method, so how do we perform logging here?

abs51295:
@msrb answered ;)

f8a_worker/workers/dependency_parser.py
abs51295: Nope, we can make it non-static :)

abs51295:
@msrb for internal scans, we still depend on workers.

PR #730, opened 2018-10-23
Introduce "go" backend for Golang ecosystem

msrb:

PR #732, opened 2018-10-23
Switch to "go" backend

msrb:


msrb:
Context: https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/730

PR #736, opened 2018-11-07
[exec_tests.sh] show locals in tracebacks

jpopelka:
https://docs.pytest.org/en/latest/usage.html#modifying-python-traceback-printing


msrb:
[test]

PR #765, opened 2019-01-03
Don't throw an exception if we process the same API request multiple times

msrb:
Fixes:
psycopg2.IntegrityError: duplicate key value violates unique constraint "api_requests_pkey"

### DIFF
--- a/f8a_worker/storages/postgres.py
+++ b/f8a_worker/storages/postgres.py
@@ -7,7 +7,7 @@ import hashlib
 import json
 from itertools import chain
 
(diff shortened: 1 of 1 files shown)


msrb:
[test]

msrb:
Ehm, no CI?

msrb:
@yzainee I believe this should fix the issue with `BookkeeperTask`. It's possible that similar fix will need to be applied in other places as well (for osio flow for example). In any case, this is hard to test outside staging/prod, so I'd suggest simply merging this PR and check in staging.

PR #768, opened 2019-01-04
Fix bucket name for S3IssuesPRs storage

msrb:
https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html

Since March 1, 2018, Amazon S3 no longer supports creating bucket names
that contain uppercase letters or underscores.


animuk:
LGTM

PR #771, opened 2019-01-07
Changed Json Format in S3

animuk:
Added tags sources, link and type


msrb:
@animuk since we modify the resulting JSON here, don't we need to update unit tests as well?

animuk:
@msrb Actually, Overall JSON structure remains same. Only modified the 'details' key object. So, test cases are still intact

PR #774, opened 2019-01-10
Fix "Path '.../extracted_jar'" does not exist" error

msrb:
Bug: https://github.com/openshiftio/openshift.io/issues/4700


f8a_worker/process.py
```diff
@@ -208,6 +208,12 @@ class Archive(object):
     @staticmethod
     def extract(target, dest):
         """Detect archive type and extracts it."""
+        # Make sure that the destination directory exists
+        try:
+            Path(dest).mkdir(mode=0o777, parents=True)
```
msrb: `exists_ok` param in not available in Python 3.4

msrb:
Moved: https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/775

PR #775, opened 2019-01-11
Fix bugs: gh#4056 and gh#4700

msrb:
Fixes: https://github.com/openshiftio/openshift.io/issues/4056 and https://github.com/openshiftio/openshift.io/issues/4700


msrb:
[test]

PR #776, opened 2019-01-11
Drop RubyGems bits

msrb:
RubyGems support was added very early on, during CuCoS days. Nobody's
been using it/keeping it up-to-date since then.


msrb:
[test]

PR #788, opened 2019-03-05
Add check for Mac OS

ravsa:


hack/generate-db-migrations.sh
```diff
@@ -37,8 +37,14 @@ done
...
-#for MAC docker run -t -v `pwd`:/bayesian \
-docker run -t -v "$(readlink -f "${THISDIR}/.."):/f8a_worker:z" \
+if [ "$(uname)" == "Darwin" ]; then
+    #for MAC docker
+    vol_mount="`pwd`:/f8a_worker:z"
+else
+    vol_mount="$(readlink -f "${THISDIR}/.."):/f8a_worker:z"
+fi
+
+docker run -t -v $vol_mount \
```
msrb: I _think_ this will break for volume paths containing spaces.

msrb: I.e. you want to retain the `" "`

ravsa: Nice catch, Thanks

ravsa:
[test]

hack/generate-db-migrations.sh
msrb: I meant doing something like this:
`docker run -t -v "$vol_mount"`

:)

PR #790, opened 2019-03-15
Make sure "declared_licenses" field is None if there is no license

msrb:
Previously it could have been "" (empty string).


msrb:
cc @tisnik

PR #795, opened 2019-03-18
Make sure description in NPM package is always string

msrb:
npm:armlet:2.3.0 have description as an array of strings.


f8a_worker/data_normalizer/javascript.py
```diff
@@ -176,6 +177,16 @@ def _transform_declared_licenses_list(self, k, value):
                     licenses.append(l["name"])
         self._data[k] = licenses
 
+    def _transform_description(self):
+        key = 'description'
+        value = self._data[key]
+        if isinstance(value, str):
+            return
+        elif isinstance(value, list):
```
tisnik: maybe add a tuple as well, just for sure?

msrb:
Added tuple to the mix :wink:

PR #824, opened 2019-04-17
fix: github_worker

sunilk747:


tests/test_utils.py
```diff
@@ -187,6 +187,7 @@ class TestParseGHRepo:
         'github.com/foo/bar.git',
         'www.github.com/foo/bar',
         'http://github.com/foo/bar',
+        'https://github.com/foo/bar/bar'
```
msrb: What about `https://github.com/foo/bar/bar/anotherbar`? :P

sunilk747: @msrb I guess now it's good to handle any such cases. It will only take into consideration the first two elements which is `org+reponame`

msrb:
Let's merge https://github.com/fabric8-analytics/fabric8-analytics-worker/pull/822 first.

msrb:
[test]

msrb:
@sunilk747 please rebase and LGTM :wink:

PR #846, opened 2019-05-21
change date format

sunilk747:


yzainee-zz (commented):
This would cause a problem while quering graph db. We will always have to modify the date format. As discussed, make this change in the UI facing side and not in the db

sunilk747:
Not needed anymore. Changed the format in backbone.

PR #871, opened 2019-10-11
changes to prevent pvt ingestion

yzainee:
This PR would prevent any pvt pkgs from getting into the ingestion flow.

### DIFF
--- a/f8a_worker/workers/init_analysis_flow.py
+++ b/f8a_worker/workers/init_analysis_flow.py
@@ -12,6 +12,8 @@ from f8a_worker.base import BaseTask
 from f8a_worker.process import IndianaJones, MavenCoordinates
 from f8a_worker.models import Analysis, EcosystemBackend, Ecosystem, Version, Package
 from f8a_worker.utils import normalize_package_name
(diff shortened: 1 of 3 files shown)


f8a_worker/workers/init_analysis_flow.py
```diff
@@ -37,6 +39,15 @@ def execute(self, arguments):
         # make sure we store package name in its normalized form
         arguments['name'] = normalize_package_name(ecosystem.backend.name, arguments['name'])
 
+        # Dont try ingestion for private packages
```
msrb: Which ecosystems were problematic? We have some code that should deal with non-existent versions in https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/master/f8a_worker/process.py#L349-L361 (this one's for NPM).

So maybe there is a bug somewhere?

yzainee: this is just to prevent any pvt package from ingestion. we dont even go to the extent of checking the version

yzainee: Also, having this later in the flow, the queue gets cluttered

PR #878, opened 2019-12-04
Removed raising exception for unsuccessful scancode execution

samuzzal-choudhury:
# Description

Fix for Jira issue: https://jira.coreos.com/browse/APPAI-949 (bayesian-worker-ingestion error - FatalTaskError: Error (137) during running command '/opt/scancode-toolkit/scancode)

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### DIFF
--- a/f8a_worker/workers/license.py
+++ b/f8a_worker/workers/license.py
@@ -9,7 +9,7 @@ from f8a_worker.base import BaseTask
 from f8a_worker.schemas import SchemaRef
 from f8a_worker.object_cache import ObjectCache
 from f8a_worker.defaults import configuration
-from selinon import FatalTaskError, StoragePool
+from selinon import StoragePool
 
 
 class LicenseCheckTask(BaseTask):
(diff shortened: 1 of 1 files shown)


f8a_worker/workers/license.py
```diff
@@ -112,4 +111,9 @@ def execute(self, arguments):
             cache_path = ObjectCache.get_from_dict(arguments).get_extracted_source_tarball()
 
         result_data = self.run_scancode(cache_path)
-        return result_data
+        if 'error' in result_data:
+            self.log.error ("Error (%s) during running command %s: %r" %
```
yzainee: This will again show in sentry as an error during the ingestion flow. if we are planning to by pass this, i think we should change it to info

PR #883, opened 2020-02-13
disk quota exceeded: clearing mech for .npm folder

yzainee:


Dockerfile
```diff
@@ -3,6 +3,8 @@ FROM quay.io/openshiftio/fabric8-analytics-f8a-worker-base:aa8893d
 ENV LANG=en_US.UTF-8 \
     # place where to download & unpack artifacts
     WORKER_DATA_DIR='/var/lib/f8a_worker/worker_data' \
+    # place for npm artifacts
+    NPM_DATA_DIR='/workdir/.npm' \
```
arajkumar: You may not need to have a separate env for this. By default `npm` stores it's cache in $HOME/.npm directory. In this case `HOME` points to `/workdir`. So, `shutil.rmtree(path.join(os.environ['HOME'], '.npm'))` should be fine.

yzainee: we dont specifically add anything in the .npm folder. its downloading the artifacts implicitly when the npm -registry and other commands are executed during the analysis phase

arajkumar: Right. Instead of having a separate env variable, you can derive the path of `.npm` directory based on `HOME` env variable.

yzainee: @arajkumar PTAL at the changed files

f8a_worker/workers/init_analysis_flow.py
```diff
@@ -104,6 +105,8 @@ def execute(self, arguments):
         finally:
             # always clean up cache
             shutil.rmtree(cache_path)
+            if arguments['ecosystem'] == "npm":
```
arajkumar: I just gone through the flow, I feel like instead of being specific to ecosystem, deleting `source_tarball_path` should do the needful. Also I see that `IndianaJones.fetch_artifact` called in couple of places in `init_analysis_flow.py` for maven ecosystem, should we clear those as well?

yzainee: tar balls are stored in s3 not in local memory. They are extracted and worked upon. 
As for the indiana jones call, there are a few weird steps everywhere during the ingestion procedure. The call here we need to take is, do we make any changes in a working stable platform? My suggestion at the moment would be 'no'.

arajkumar: `source_tarball_path` variable holds a path to tarball which has to be put into S3. So deleting `source_tarball_path` after putting into S3 is as same as deleting `~/.npm`. Probable you can check that in your devcluster if you have the setup.

yzainee: sample: source_tarball_path -> /var/lib/f8a_worker/worker_data/tmpbrq6er3j/package.tgz location where we store the tarball. And then after processing, we delete all the contents which is inside /var/lib/f8a_worker/worker_data . So in the end, we dont have any tarball etc in our disk or local memory

msrb:
Note single worker pod can be processing multiple NPM tasks at the same time. Simply rmrf-ing $HOME/.npm could cause other tasks to fail, as there is a race condition now.

PR #898, opened 2020-06-19

jparsai:
# Description

Fixes # APPAI-1302

## Type of change
Also modified existing code for passing linter test

### DIFF
--- a/f8a_worker/storages/postgres.py
+++ b/f8a_worker/storages/postgres.py
@@ -17,7 +17,7 @@ from f8a_worker.models import Analysis, Ecosystem, Package, Version, WorkerResul
 from f8a_worker.utils import MavenCoordinates
 
 from .postgres_base import PostgresBase
-
+import string
 
 Base = declarative_base()
 
@@ -45,7 +45,7 @@ class BayesianPostgres(PostgresBase):
             started_at=result.get('_audit', {}).get('started_at') if result else None,
             ended_at=result.get('_audit', {}).get('ended_at') if result else None,
             analysis_id=node_args.get('document_id') if isinstance(node_args, dict) else None,
-            task_result=result,
+            task_result=validate_utf_json(result),
             error=error,
             external_request_id=(node_args.get('external_request_id')
                                  if isinstance(node_args, dict) else None)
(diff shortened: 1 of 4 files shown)


f8a_worker/storages/postgres.py
```diff
@@ -278,3 +278,17 @@ def get_analysed_versions(ecosystem, package):
...
+def validate_utf_json(result):
+    """Remove non UTF-8 characters from description if present.
+
+    :param result: dict, Dictionary having details of task
+    return: a dictionary received as input having non UTF characters removed if present
+    """
+    if 'details' in result and isinstance(result['details'], list) \
+            and len(result['details']) > 0 and 'description' in result['details'][0]:
+        result['details'][0]['description'] = ''.join(
+            x for x in result['details'][0]['description']
```
yzainee, line 292: i dont understand this step. there are ways in python to convert unicode to utf8
1. your_str.decode('utf-8', 'ignore') or
2. unicode(your_str, "utf-8")

jparsai: Yes, other functions are also there but seems this approach gives us better output as it keeps only printable characters.
unicode() function was there in Python 2 but it was replaced in Python 3 with str.
Please have a look on output using different functions:

#Input string
test_str = '��C\x00h\x00a\x00n\x00g\x00e\x00 \x001\x00'

1) ''.join(x for x in s if x in string.printable)
#Output = 'Change 1'

2) test_str.encode('utf-8','ignore').decode('utf-8','ignore')
#Output = '��C\x00h\x00a\x00n\x00g\x00e\x00 \x001\x00'

3) str(test_str.encode('utf-8'))
#Output = "b'\\xef\\xbf\\xbd\\xef\\xbf\\xbdC\\x00h\\x00a\\x00n\\x00g\\x00e\\x00 \\x001\\x00'"

yzainee: why are you doing encode and decode in option number 2?

jparsai: It is because str.encode() gives us Bytes, so we need convert it into String, for that we can use decode() as given in option 2 or str() as given in option 3.

Issue #121, opened 2017-07-17
Dependency snapshot for pypi/django-timezone-field/2.0 fails with error ValueError: Package django not found


fridex:
```python
{"node_args": {"_audit": {"started_at": "2017-07-17T08:39:28.536684", "version": "v1", "ended_at": "2017-07-17T08:39:28.594353"}, "force": true, "version": "2.0", "document_id": 1113522, "force_graph_sync": false, "ecosystem": "pypi", "_release": "pypi:django-timezone-field:2.0", "name": "django-timezone-field", "recursive_limit": 0}, "task_name": "dependency_snapshot", "record_id": null, "error_traceback": "  File \"/usr/lib/python3.4/site-packages/selinon/selinonTaskEnvelope.py\", line 114, in run\n    result = task.run(node_args)\n  File \"/usr/lib/python3.4/site-packages/f8a_worker/base.py\", line 38, in run\n    result = self.execute(node_args)\n  File \"/usr/lib/python3.4/site-packages/f8a_worker/workers/dependency_snapshot.py\", line 94, in execute\n    resolved = self._resolve_dependency(ecosystem, dep)\n  File \"/usr/lib/python3.4/site-packages/f8a_worker/workers/dependency_snapshot.py\", line 64, in _resolve_dependency\n    pkgspec = solver.solve([dep])\n  File \"/usr/lib/python3.4/site-packages/f8a_worker/solver.py\", line 585, in solve\n    name, releases = self.release_fetcher.fetch_releases(dep.name)\n  File \"/usr/lib/python3.4/site-packages/f8a_worker/solver.py\", line 148, in fetch_releases\n    return self.fetch_releases(self._search_package_name(package))\n  File \"/usr/lib/python3.4/site-packages/f8a_worker/solver.py\", line 135, in _search_package_name\n    raise ValueError(\"Package {} not found\".format(package))\n", "storage_task_name": "dependency_snapshot", "flow_name": "bayesianAnalysisFlow", "error_type": "<class 'ValueError'>", "event": "STORAGE_STORE_ERROR", "error_value": "Package django not found", "storage_name": "BayesianPostgres", "task_id": "c26d5b8d-f821-4620-b172-07b6b40cea66"}
```

PR #899, opened 2020-06-25

jparsai:
# Description

Fixes # APPAI-1302

## Type of change
Handling None type Objects, also put try-except blocks, in case of failure doing nothing and returning original data back, using print() as linter tests dont allow empty catch blocks.

### DIFF
--- a/f8a_worker/storages/postgres.py
+++ b/f8a_worker/storages/postgres.py
@@ -17,7 +17,9 @@ from f8a_worker.models import Analysis, Ecosystem, Package, Version, WorkerResul
 from f8a_worker.utils import MavenCoordinates
 
 from .postgres_base import PostgresBase
-import string
+import logging
+
+logger = logging.getLogger(__name__)
 
 Base = declarative_base()
 
@@ -45,7 +47,7 @@ class BayesianPostgres(PostgresBase):
             started_at=result.get('_audit', {}).get('started_at') if result else None,
             ended_at=result.get('_audit', {}).get('ended_at') if result else None,
             analysis_id=node_args.get('document_id') if isinstance(node_args, dict) else None,
-            task_result=validate_utf_json(result),
+            task_result=result,
             error=error,
             external_request_id=(node_args.get('external_request_id')
                                  if isinstance(node_args, dict) else None)
@@ -278,17 +280,3 @@ class BayesianPostgres(PostgresBase):
         except SQLAlchemyError:
             PostgresBase.session.rollback()
             raise
-
-
-def validate_utf_json(result):
-    """Remove non UTF-8 characters from description if present.
-
-    :param result: dict, Dictionary having details of task
-    return: a dictionary received as input having non UTF characters removed if present
-    """
-    if 'details' in result and isinstance(result['details'], list) \
-            and len(result['details']) > 0 and 'description' in result['details'][0]:
-        result['details'][0]['description'] = ''.join(
-            x for x in result['details'][0]['description']
-            if x in string.printable)
-    return result
--- a/f8a_worker/workers/init_package_flow.py
+++ b/f8a_worker/workers/init_package_flow.py
@@ -9,6 +9,9 @@ from f8a_worker.errors import NotABugFatalTaskError
 from f8a_worker.base import BaseTask
 from f8a_worker.models import Ecosystem, Package, Upstream, PackageAnalysis
 from urllib.parse import urlparse
+import logging
+
+logger = logging.getLogger(__name__)
 
 
 class InitPackageFlow(BaseTask):
@@ -90,7 +93,7 @@ class InitPackageFlow(BaseTask):
                           "upstream URL '%s'", package.ecosystem.name, package.name, url)
             new_upstream = Upstream(
                 package_id=package.id,
-                url=validate_url(url),
+                url=_validate_url(url),
                 updated_at=None,
                 deactivated_at=None,
                 added_at=now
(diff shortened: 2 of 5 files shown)


f8a_worker/workers/init_package_flow.py
```diff
@@ -179,5 +179,6 @@ def validate_url(url):
             return url
         else:
             return ''
-    except ValueError:
+    except Exception:
+        print()
```
yzainee, line 183: print the exception. at least it will be useful. use logger instead

f8a_worker/storages/postgres.py
```diff
@@ -286,9 +286,13 @@ def validate_utf_json(result):
...
-    return result
+    try:
+        if result is not None and 'details' in result and isinstance(result['details'], list) \
+                and len(result['details']) > 0 and 'description' in result['details'][0]:
+            result['details'][0]['description'] = ''.join(
+                x for x in result['details'][0]['description']
+                if x in string.printable)
+        return result
+    except Exception:
+        print()
```
yzainee, line 297: same goes here as well

jparsai:
@yzainee , @arajkumar I have made changes as you suggested and requested a re-review, Please review again.

f8a_worker/storages/postgres.py
```diff
@@ -286,9 +289,13 @@ def validate_utf_json(result):
...
     """
-    if 'details' in result and isinstance(result['details'], list) \
-            and len(result['details']) > 0 and 'description' in result['details'][0]:
-        result['details'][0]['description'] = ''.join(
-            x for x in result['details'][0]['description']
-            if x in string.printable)
-    return result
+    try:
+        if result is not None and 'details' in result and isinstance(result['details'], list) \
+                and len(result['details']) > 0 and 'description' in result['details'][0]:
```
arajkumar, line 294: I'm finding little difficulty in following the purpose of the condition. After so many criteria, we are checking whether the `details` field has printable chars. 

If the criteria fails we are returning the result as such. Does it mean only specific worker type sets `details` field? If that is the case, can't we validate the `details` in that particular worker itself?

f8a_worker/storages/postgres.py
```diff
@@ -286,9 +289,13 @@ def validate_utf_json(result):
...
+            result['details'][0]['description'] = ''.join(
+                x for x in result['details'][0]['description']
+                if x in string.printable)
+        return result
+    except Exception as e:
```
arajkumar, line 299: Also I'm little hesitant to use base Exception class to handle it. It would mask the real issue which may need our attention. Can you use specific types of Exceptions? (possible exception classes could be KeyError, ValueError)

f8a_worker/storages/postgres.py
jparsai: Yes @arajkumar, only some places only description is present, as this is the common place so I have done it here.
Allow me some time to list down those places and see if I can write this checks there.

f8a_worker/storages/postgres.py
jparsai: Okay @arajkumar I will keep multiple catch blocks for KeyError and ValueError but just to be on safe side I will also keep base Exception as well.

arajkumar: @jparsai I don't recommend to use base Exception class unless there is a valid reason for that to have.

arajkumar: You can do the following instead of multiple blocks,
```
except (KeyError, ValueError) as e:
  ...
```

f8a_worker/workers/mercator.py
```diff
@@ -271,3 +275,22 @@ def get_outermost_items(list_):
             sorted_list = [i for i in sorted_list if
                            len(i['path'].split(os.path.sep)) == outermost_len]
         return sorted_list
+
+
+def validate_utf_json(result):
```
arajkumar, line 280: This can be a private function as nobody else accesses it from outside.

f8a_worker/workers/mercator.py
```diff
@@ -271,3 +275,22 @@ def get_outermost_items(list_):
...
+    """Remove non UTF-8 characters from description if present.
+
+    :param result: dict, Dictionary having details of task
+    return: a dictionary received as input having non UTF characters removed if present
+    """
+    try:
+        if result is not None and 'details' in result and isinstance(result['details'], list):
```
arajkumar, line 287: Since the UTF 8 validation is specific to mercator, it is quite evident that `details` field will always exist(See line #121) and you can skip`if result is not None and 'details' in result and isinstance(result['details'], list):` and exception handling too.
```suggestion

```

f8a_worker/workers/mercator.py
```diff
@@ -271,3 +275,22 @@ def get_outermost_items(list_):
...
+            for i in range(len(result['details'])):
+                if 'description' in result['details'][i]:
+                    result['details'][i]['description'] = ''.join(
+                        x for x in result['details'][i]['description']
+                        if x in string.printable)
```
arajkumar, line 292: List iteration shall be simplified as below,

```suggestion
            for detail in result['details']:
                if 'description' in detail:
                    detail['description'] = ''.join(
                        x for x in detail['description']
                        if x in string.printable)
```

f8a_worker/workers/mercator.py
```diff
@@ -271,3 +275,22 @@ def get_outermost_items(list_):
...
+        return result
+    except (KeyError, ValueError) as e:
+        logger.error("Invalid dictionary data: {}".format(str(e)))
+        return result
```
arajkumar, line 296: `try..catch` may not be required as described above.
```suggestion
```

f8a_worker/workers/mercator.py
yzainee: mercator is used for the pkg analysis. Its not used for stack reports

PR #900, opened 2020-06-29

jparsai:
# Description

Fixes # APPAI-1302

## Type of change
For some packages description field is present but value is None, hence adding that check in handler.


jparsai:
@arajkumar Please review this one change in the condition.

arajkumar:
@jparsai Please update the test to cover this scenario.

arajkumar:
@yzainee Any concerns?

PR #903, opened 2020-07-03
fix: Use custom selinon with patch to drain message if it is old

jparsai:
# Description
Updating command to install selinon from binary is rhel dockerfile 

## Fixes
[APPAI-1359]


yzainee:
[test]

arajkumar:
[test]

PR #911, opened 2020-07-30
fix: Monkey patch selinon to drain out old messages

jparsai:
# Description
- Used monkey patching to ingest the logic in Selinon to drain out messages which are stuck in queue for more than configured time.
- Installed Selinon from PyPi repository instead of binary file.
- Introduced a new Env variable to configure time limit for removing messages from Selinon.

## Issues
APPAI-1359


f8a_worker/start.py
```diff
@@ -7,6 +7,52 @@
...
+
+        if no_of_hours >= dispatcher_time_out_in_hrs:
+            exc = NotABugFatalTaskError("Flow is running for {} Hours. "
+                                        "It is being stopped forcefully. "
+                                        "Flow information: {}".format(no_of_hours, flow_info))
+            raise self.retry(max_retries=0, exc=exc)
+    else:
+        # If message is arrived for the first time, then put current time in node arguments
+        # and consider it as starting time of the flow.
+        node_args['flow_start_time'] = str(datetime.now())
```
arajkumar, line 55: Move these piece of code into monkey_patch.py module and it should just expose `patch` method.

f8a_worker/start.py
```diff
@@ -7,6 +7,52 @@
 from f8a_worker.setup_celery import init_celery, init_selinon
 import raven
 from raven.contrib.celery import register_signal, register_logger_signal
+from selinon import Dispatcher
+from datetime import datetime
+import json
+from selinon.errors import SelinonException
+
+
+class NotABugFatalTaskError(SelinonException):
```
arajkumar, line 16: Can't this be imported from selinon itself?

`from selinon.errors import NotABugFatalTaskError`

openshift/template.yaml
```diff
@@ -167,6 +167,8 @@ objects:
               configMapKeyRef:
                 name: bayesian-config
                 key: auth-url
+          - name: DISPATCHER_TIME_OUT_IN_HRS
```
arajkumar, line 170: It is little difficult to graph, can you find a good name? SQS_MSG_LIFETIME?

f8a_worker/start.py
jparsai: @arajkumar this Exception is not is Selinon,  in worker also we have created in f8a_worker/errors.py

jparsai: https://github.com/selinon/selinon/blob/master/selinon/errors.py
This Exception is not present in Selinon, a custom NotABugFatalTaskError is created because we are using same in workers as well and I think it is ignored by sentry in reporting.

f8a_worker/monkey_patch.py
```diff
@@ -0,0 +1,63 @@
...
+        """Make sure flow errors capture final state of the flow.
+
+        :param state: final flow details
+        """
+        super().__init__(json.dumps(state))
+
+    @property
+    def state(self):
+        """Get structured flow details."""
+        return json.loads(str(self))
```
arajkumar, line 23: If it is already defined in f8aworker.errors, why not reuse the same here?

f8a_worker/monkey_patch.py
```diff
@@ -0,0 +1,63 @@
...
+
+        if no_of_hours >= dispatcher_time_out_in_hrs:
+            exc = NotABugFatalTaskError("Flow is running for {} Hours. "
+                                        "It is being stopped forcefully. "
+                                        "Flow information: {}".format(no_of_hours, flow_info))
+            raise self.retry(max_retries=0, exc=exc)
+    else:
+        # If message is arrived for the first time, then put current time in node arguments
+        # and consider it as starting time of the flow.
+        node_args['flow_start_time'] = str(datetime.now())
```
arajkumar, line 50: Just deal with timestamp rather than string, (using time module)

```suggestion
   flow_start_time = flow_info.get('node_args', {}).get('flow_start_time', 0)
   if flow_start_time > 0:
        now = time.time()
        if  now - flow_start_time > _SQS_MSG_LIFETIME_IN_SEC:
            exc = NotABugFatalTaskError("Flow '{}' Timeout: started at {} now {}. ".format(flow_info, flow_start_time, now)
            raise self.retry(max_retries=0, exc=exc)
    else:
        # New flow, add time stamp.
        node_args['flow_start_time'] = time.time()
```

f8a_worker/monkey_patch.py
```diff
@@ -0,0 +1,63 @@
...
+    node_args = flow_info['node_args']
+    if node_args is not None and 'flow_start_time' in node_args \
+            and node_args['flow_start_time'] is not None:
+        flow_start_time = datetime.strptime(node_args['flow_start_time'], '%Y-%m-%d %H:%M:%S.%f')
+        current_time = datetime.now()
+        time_diff = current_time - flow_start_time
+        no_of_hours = time_diff.days * 24 + time_diff.seconds // 3600
+
+        node_args['no_of_hours'] = no_of_hours
+        dispatcher_time_out_in_hrs = int(os.environ.get('SQS_MSG_LIFETIME', '24'))
```
arajkumar, line 40: Convert it as global package private constant and convert the hours to seconds to avoid repeated conversion during the condition testing.
```suggestion
     _SQS_MSG_LIFETIME_IN_SEC   = int(os.environ.get('SQS_MSG_LIFETIME', '24')) * 60 * 60
```

jparsai:
@yzainee Please review the changes

f8a_worker/monkey_patch.py
```diff
@@ -0,0 +1,41 @@
...
+        node_args['flow_start_time'] = time.time()
+
+
+def patch(self):
+    """Monkey Patching "Dispatcher.migrate_message" function to modify code at runtime."""
+    original_migrate_message = Dispatcher.migrate_message
+
+    def patched_migrate_message(self, flow_info):
+        res = original_migrate_message(self, flow_info)
+        # Adding patch to throw error on message is too old
```
yzainee, line 37: throw error on message is too old -> throw error if the message is older than SQS_MSG_LIFETIME

arajkumar:
@jparsai Why did you squash by yourself? There is an option in Github to squash during merge.

PR #915, opened 2020-08-06
fix: Define SQS_MSG_LIFETIME as oc template parameter

jparsai:
# Description

As CI deployment job does not accept value of a environment variable as a Numeric type, making the value a String object.

### DIFF
--- a/openshift/template.yaml
+++ b/openshift/template.yaml
@@ -168,7 +168,7 @@ objects:
                 name: bayesian-config
                 key: auth-url
           - name: SQS_MSG_LIFETIME
-            value: "24"
+            value: "${SQS_MSG_LIFETIME}"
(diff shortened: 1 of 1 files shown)


openshift/template.yaml
```diff
@@ -168,7 +168,7 @@ objects:
                 name: bayesian-config
                 key: auth-url
           - name: SQS_MSG_LIFETIME
-            value: 24
+            value: "24"
```
arajkumar, line 171: This should be template parameterised. Look at the parameter section & do accordingly.

jparsai:
[test]

arajkumar:
[test]

jparsai:
[test]

jparsai:
Build was failing because NPM registry was down.
https://registry.npmjs.org/@slicemenice/event-utils

arajkumar:
>Build was failing because NPM registry was down.
 You should mock out. it is not good for an unit test to deal with a fragile part.

PR #921, opened 2020-11-11
Migrate CI to GitHub Actions

sawood14012:
# Description

Added/updated Scripts to enable Github actions

## Type of change

Please delete options that are not relevant.

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)


sawood14012:
video shows how the build-image comment will be edited - to make sure there are no redundant comments 
https://user-images.githubusercontent.com/18240985/105129534-1fce2880-5a9a-11eb-8628-2d40c79e5b51.mov

qa/runtest.sh
```diff
@@ -95,11 +95,12 @@ docker run -d \
 S3_CONTAINER_IP=$(docker inspect --format "{{.NetworkSettings.Networks.${DOCKER_NETWORK}.IPAddress}}" ${TESTS3_CONTAINER_NAME})
 S3_ENDPOINT_URL="http://${S3_CONTAINER_IP}:33000"
 
+mkdir shared
+
 echo "Starting test suite"
-docker run -t \
-  -v "${here}:/f8a_worker:ro,Z" \
+docker run -v "$PWD/shared:/tmp/shared:rw,Z" -t \
```
arajkumar, line 101: Could you just mount the $PWD, so that you can avoid dealing with extra dir?

arajkumar (changes requested):

qa/runtest.sh
sawood14012: okay

arajkumar: @sawood14012 , This won't work when REBUILD is set to 0. `  -v "${here}:/f8a_worker:ro,Z"` is a must, my argument is that should be sufficient if it is mounted in rw. Let me check from my end to understand the issue which you are facing.

sawood14012: okay

arajkumar: You can use the following changes. 
https://github.com/arajkumar/fabric8-analytics-worker/commit/d9d3abbddca828d8f2bb44d1fe4765a8087c6431

You wouldn't have seen coverage.xml which is written in root dir because it is in .gitignore.

arajkumar (approved):
lgtm

PR #924, opened 2020-11-24
feat: Collect more Github statistics

jparsai:
# Description
Added logic to collect data for missing github details for both new and old flows.

## Issue
APPAI - 1608

## Type of changes
1. Written common logic to be called from old and new ingestion flows.
2. Removed schema validation from old Github class. The version of schema has been changed and the fields added were part of old schema because of this, it is causing validation error. Since we are already creating JSON in format required by Data Importer we can avoid validation of schema in worker.

### DIFF
--- a/f8a_worker/utils.py
+++ b/f8a_worker/utils.py
@@ -632,6 +632,7 @@ def store_data_to_s3(arguments, s3, result):
         logger.error(e)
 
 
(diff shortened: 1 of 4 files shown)


jparsai:
[test]

yzainee (approved):
Im fine with the changes.. just worried about the number of gh calls that we are doing

jparsai:
@arajkumar do you have any way around?

PR #927, opened 2020-12-17
chore : Pull minio from quay instead of docker 

sawood14012:
# Description

Use minio from quay rather than docker hub to avoid pull limit errors


dgpatelgit (approved):
lgtm

PR #932, opened 2020-12-28
chore: Dependency upgrade

jparsai:
# Type of changes
- [x] Updated Utils version to latest.
- [x] Pinned celery version to 4.x as 5.x has deprecated some functions, changes to support 5.x can be done later.


yzainee (approved):
LGTM

PR #937, opened 2021-01-11
fix: Persist uuid into user_id column of api_requests table.

rafiu007:
Now we can store uuid from CA calls into the api_requests table.

### DIFF
--- a/f8a_worker/storages/postgres.py
+++ b/f8a_worker/storages/postgres.py
@@ -213,7 +213,7 @@ class BayesianPostgres(PostgresBase):
         s3 = StoragePool.get_connected_storage('S3UserProfileStore')
         s3.store_in_bucket(content)
 
-    def store_api_requests(self, external_request_id, data, dep_data):
+    def store_api_requests(self, external_request_id, uuid_data, data, dep_data):
         """Get result of previously scheduled analysis.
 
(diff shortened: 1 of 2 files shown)


f8a_worker/storages/postgres.py
```diff
@@ -213,7 +213,7 @@ def store_in_bucket(content):
         s3 = StoragePool.get_connected_storage('S3UserProfileStore')
         s3.store_in_bucket(content)
 
-    def store_api_requests(self, external_request_id, data, dep_data):
+    def store_api_requests(self, external_request_id, uuid_data, data, dep_data):
```
prashbnair, line 216: There is scope for clean up in this table, a lot of the information that we store is not useful. A lot of data is none and I am not sure why those columns exist. It will be good to store the ecosystem that is used in plain text. I also believe that it useful to store the stack itself.

cc @arajkumar @yzainee

prashbnair (changes requested):

prashbnair:
@arajkumar @yzainee can we decide what we want to track for the CA calls, my feeling is that the ecosystem and stack itself is useful information.

rafiu007:
The team has decided to go with a new table for storing data from CA calls.

PR #939, opened 2021-01-15
fix: Add new table `component_analysyes_requests`.

rafiu007:
Added new table for storing data from CA calls.Look at the doc below to see specs.
https://docs.google.com/document/d/1JxEYgMSVHPBevHvjdvPX_cRUeUO0Be-0nGEnOXvRNXc/edit?usp=sharing


f8a_worker/models.py
```diff
@@ -428,3 +428,18 @@ class OSIORegisteredRepos(Base):
...
     last_scanned_at = Column(DateTime)
+
+
+class ComponentAnalysesRequests(Base):
+    """Table for CA request."""
+
+    __tablename__ = "component_analyses_requests"
+
+    component_analyses_request_id = Column(String(64), primary_key=True)
+    request_id = Column(String(64), nullable=True)
```
prashbnair, line 439: This is should be a UUID. It will be generated for each request.

f8a_worker/models.py
```diff
@@ -428,3 +428,18 @@ class OSIORegisteredRepos(Base):
...
     email_ids = Column(String(255), nullable=False)
     last_scanned_at = Column(DateTime)
+
+
+class ComponentAnalysesRequests(Base):
+    """Table for CA request."""
+
+    __tablename__ = "component_analyses_requests"
+
+    component_analyses_request_id = Column(String(64), primary_key=True)
```
prashbnair, line 438: Should the generated primary key be a string or a number?

prashbnair (changes requested):

f8a_worker/models.py
rafiu007: @prashbnair  Keeping it  as a string seems more flexible.the same has been done for other tables as well.

rafiu007: Done.

f8a_worker/models.py
rafiu007: Done

prashbnair (approved):

PR #945, opened 2021-02-02
fix: Modify `ComponentApiFlow` to store data from CA post call into RDS

rafiu007:
A new fucntion has been added in component api flow to store data inside component analyses request table.

### DIFF
--- a/f8a_worker/workers/bookkeeper.py
+++ b/f8a_worker/workers/bookkeeper.py
@@ -77,5 +77,8 @@ class BookkeeperTask(BaseTask):
             self.store_user_node(arguments, aggregated)
 
         postgres = StoragePool.get_connected_storage('BayesianPostgres')
-        postgres.store_api_requests(arguments.get('external_request_id'),
-                                    arguments.get('data'), aggregated)
+        if arguments.get('api_name') == 'component_analyses_post':
+            postgres.store_api_requests_post(arguments)
+        else:
+            postgres.store_api_requests(arguments.get('external_request_id'),
+                                        arguments.get('data'), aggregated)
--- a/f8a_worker/workers/bookkeeper.py
+++ b/f8a_worker/workers/bookkeeper.py
@@ -71,13 +71,13 @@ class BookkeeperTask(BaseTask):
         self._strict_assert(arguments.get('data'))
 
         aggregated = ''
-        if arguments.get('api_name') == 'stack_analyses' and \
+        if arguments['data'].get('api_name') == 'stack_analyses' and \
            'email' in arguments['data'].get('user_profile', {}):
             aggregated = self.parent_task_result('GraphAggregatorTask')
             self.store_user_node(arguments, aggregated)
 
         postgres = StoragePool.get_connected_storage('BayesianPostgres')
-        if arguments.get('api_name') == 'component_analyses_post':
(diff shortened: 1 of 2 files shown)


f8a_worker/workers/bookkeeper.py
```diff
@@ -71,11 +71,14 @@ def execute(self, arguments):
...
            'email' in arguments['data'].get('user_profile', {}):
             aggregated = self.parent_task_result('GraphAggregatorTask')
             self.store_user_node(arguments, aggregated)
 
         postgres = StoragePool.get_connected_storage('BayesianPostgres')
-        postgres.store_api_requests(arguments.get('external_request_id'),
-                                    arguments.get('data'), aggregated)
+        if arguments.get('api_name') == 'component_analyses_post':
+            postgres.store_api_requests_post(arguments)
+        else:
```
prashbnair, line 82: I am not sure if this flow is still being used, if not can we delete it.

f8a_worker/workers/bookkeeper.py
```diff
@@ -71,11 +71,14 @@ def execute(self, arguments):
         self._strict_assert(arguments.get('data'))
 
         aggregated = ''
-        if arguments['data'].get('api_name') == 'stack_analyses' and \
+        if arguments.get('api_name') == 'stack_analyses' and \
```
prashbnair, line 74: Why have you changed the code for stack analyses. Does it mean that is not working currently in prod.

prashbnair (changes requested):

f8a_worker/workers/bookkeeper.py
rafiu007: @prashbnair   I am not sure whats happening in prod.But that was like a syntax error that I saw in the worker. "string does not have get() attribute".

f8a_worker/workers/bookkeeper.py
rafiu007: if we get Ca get calls it will be used.Otherwise it wont.

f8a_worker/workers/bookkeeper.py
rafiu007: resolved.

f8a_worker/workers/bookkeeper.py
rafiu007: resolved

f8a_worker/workers/bookkeeper.py
```diff
@@ -77,5 +77,5 @@ def execute(self, arguments):
             self.store_user_node(arguments, aggregated)
 
         postgres = StoragePool.get_connected_storage('BayesianPostgres')
-        postgres.store_api_requests(arguments.get('external_request_id'),
-                                    arguments.get('data'), aggregated)
+        if arguments['data'].get('api_name') == 'component_analyses_post':
```
prashbnair, line 80: Why is this check required?

prashbnair (changes requested):

f8a_worker/workers/bookkeeper.py
rafiu007: @prashbnair I have kept the check as there is a check for sa bookkeeping as well.It will also help when this goes into stage.As otherwise other bookkeeping tasks will trigger this one..

prashbnair (approved):

PR #948, opened 2021-02-11
fix: Add column source to `component_analyses_requests` table.

rafiu007:
This new column will be used to store the source of the request.


prashbnair (changes requested):
Where is the script to change the data model?

rafiu007:
> Where is the script to change the data model?

I was waiting for +1 for the column name.

f8a_worker/storages/postgres.py
```diff
@@ -231,7 +231,8 @@ def store_api_requests_post(self, arguments):
             ecosystem=arguments['data'].get('ecosystem'),
             user_agent=arguments['data'].get('user_agent'),
             stack_data=json.dumps(arguments['data'].get('packages_list')),
-            manifest_hash=arguments['data'].get('manifest_hash')
+            manifest_hash=arguments['data'].get('manifest_hash'),
+            source=arguments['data'].get('source')
```
prashbnair, line 235: Are you planning to add the redhat_commons_id later?

prashbnair: I would prefer telemetry_id.

rafiu007: Yes.first I will modify the api-server then add it here.

prashbnair:
> I was waiting for +1 for the column name.

I would prefer telemetry_id.

rafiu007:
> > I was waiting for +1 for the column name.
> 
> I would prefer telemetry_id.

Okay going ahead with that.

prashbnair (approved):
Please do not merge until you have everything ready or are willing to push this out to prod in a couple of days.

PR #950, opened 2021-02-23
Add App SRE build scripts and Dockerfiles

maorfr:
As part of https://issues.redhat.com/browse/APPSRE-3082 we are migrating the service to App SRE standards.
We are duplicating the Dockerfiles to pull and push to quay.io/app-sre and this will be cleaned up in a following PR.

Thanks!


arajkumar (approved):
lgtm

PR #951, opened 2021-02-23
add pull secret to openshift manifest

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3082

this PR adds a ServiceAccount to the deployment manifest so that the pods can pull a private image. the pull secret is currently associated to the default ServiceAccount, and that is not a good practice.


maorfr:
actually we'll probably need to do something different due to all the yaml anchoring in https://github.com/openshiftio/saas-analytics/blob/master/bay-services/worker.yaml

maorfr:
i'm going to merge this as we are currently migrating in https://coreos.slack.com/archives/C01DKUK9P1A/p1614152512012100

PR #957, opened 2021-02-28
fix: Retry mechanism modified.

jparsai:
# Description

Updated tenacity retry statements, and using it in wrapper function "execute_gh_queries" as well. If "get_gh_query_response" fails for all 5 attempts "execute_gh_queries" will retry it again.

### DIFF
--- a/f8a_worker/utils.py
+++ b/f8a_worker/utils.py
@@ -632,7 +632,8 @@ def store_data_to_s3(arguments, s3, result):
         logger.error(e)
 
 
-@tenacity.retry(reraise=True, stop=tenacity.stop_after_attempt(3), wait=tenacity.wait_fixed(1))
+@tenacity.retry(stop=tenacity.stop_after_attempt(5),
+                wait=tenacity.wait_exponential(multiplier=5, min=15, max=80))
 def get_gh_query_response(repo_name, status, type, start_date, end_date, event):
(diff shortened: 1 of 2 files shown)


f8a_worker/utils.py
```diff
@@ -632,7 +632,8 @@ def store_data_to_s3(arguments, s3, result):
         logger.error(e)
 
 
-@tenacity.retry(reraise=True, stop=tenacity.stop_after_attempt(3), wait=tenacity.wait_fixed(1))
+@tenacity.retry(stop=tenacity.stop_after_attempt(5),
+                wait=tenacity.wait_exponential(multiplier=5, min=15, max=80))
```
yzainee, line 636: whats the total time it will wait and how many retries will it do before aborting?

jparsai: It will try for 5 times and wait time will increase exponentially, it will wait for 2^x * 5 second (where x is iteration count)  between each retry starting with 15 seconds, then up to 20 seconds, then 30 seconds afterwards.

yzainee: when will it stop eventually?

jparsai: After 5 attempts

yzainee (approved):
LGTM

PR #958, opened 2021-03-03
fix: Repo name and retry mechanism

jparsai:
# Description
for some repositories Github API uses full name, and it is causing API to fail, to avoid it full name of repo is taken from metadata of repository.
Also retry mechanism is modified while calling any GH API.


yzainee (approved):
LGTM

